PR #5, opened 2017-05-16
sales order grid filters show in the bottom hidden

cipribucur:
Due to lack of priority into modules, if Magento_Sales is loaded in config.php later than taxjar module, the sales/order grid filters will be hidden under the grid, not visible before the grid.

### DIFF
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -17,6 +17,8 @@
 
 namespace Taxjar\SalesTax\Model;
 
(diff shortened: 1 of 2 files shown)


fastdivision:
Hi @cipribucur, thanks for contributing! Could you please open a new PR with [this single commit](https://github.com/taxjar/taxjar-magento2-extension/pull/5/commits/c8ecc88ea879d97fadfae26dc988325645d3c4d1) and a screenshot of the issue? Much appreciated.

PR #6, opened 2017-05-23
sales order grid filters show in the bottom hidden

cipribucur:
- if the sequence is not set, it could be the case to end up with this module in the config file before the Magento_Sales module, which is going to make  filter block in the adminhtml/sales/order grid invisible and move it after the grid.


fastdivision:
Thanks @cipribucur!

PR #7, opened 2017-05-24
configurable product id support

cipribucur:
for the configurable products export, the id should be unique to work. In the case of a configurable product, the initial id was taken from the configurable product. So if was the case where a client will by the same configurable product with different child products, the import will fail due to having the same id on both exported items. From a previous discussion i've understand that you accept id's that are not integers, so no the exported item id is : configurableProductId_simpleProductId.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "taxjar/module-taxjar",
     "description": "TaxJar Sales Tax Module for Magento 2",
     "type": "magento2-module",
-    "version": "0.7.1",
+    "version": "0.7.2",
     "license": "OSL-3.0",
     "require": {
         "magento/framework": "^100.1.0"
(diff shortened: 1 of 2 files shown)


fastdivision:
I believe this issue was fixed in a [previous commit](https://github.com/taxjar/taxjar-magento2-extension/commit/94759b63563af210edfe4484ae3071fe99b5ec9a) changing the value of `id` from `$item->getProductId()` to `$item->getItemId()`.

Our API only uses the line item `id` parameter to ensure duplicate line items aren't processed, but it's not saved or displayed in TaxJar. The item's exact product ID isn't necessary since we provide the SKU via `product_identifier`. So providing `$item->getItemId()` should do the trick for the same configurable products with different child products, since a unique ID is provided by line item.

PR #30, opened 2018-07-31
Modified logger to check for run mode either dev or otherwise only wr…

DiscipleOfOne:
…ites to log file if in dev mode, and added logging to the calculations, start success error and timeout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [1.0.3] - 2018-07-19
+- Fix "None" tax class product exemptions when syncing transactions.
+
 ## [1.0.2] - 2018-07-02
 - Fix discounts applied to shipping amounts for calculations.
 - Directly apply shipping discount to order-level `shipping` param and remove shipping discount line item when syncing transactions.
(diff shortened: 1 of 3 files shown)


Model/Logger.php
```diff
@@ -47,16 +48,48 @@ class Logger
...
+     * @var \Magento\Framework\App\Config\ScopeConfigInterface
+     */
+    protected $scopeConfig;
+
+    /**
+     * @var string
+     */
+    protected $filename = TaxjarConfig::TAXJAR_DEFAULT_LOG;
+
+
```
fastdivision: Drop additional newline

Model/Logger.php
```diff
@@ -47,16 +48,48 @@ class Logger
...
     ) {
         $this->directoryList = $directoryList;
         $this->driverFile = $driverFile;
+        $this->scopeConfig = $scopeConfig;
+        $this->storeManager = $storeManager;
+    }
+
+    /**
+     * Sets the filename to save the log to
+     * @param string $filename
```
fastdivision: Newline between description and params

Model/Logger.php
```diff
@@ -47,16 +48,48 @@ class Logger
...
+     * @return string
```
fastdivision: Returns strings or logger class?

Model/Logger.php
```diff
@@ -77,24 +110,33 @@ public function getPath()
...
-            $message = sprintf('%s%s - %s%s', PHP_EOL, $timestamp, $label, $message);
-            $this->driverFile->filePutContents($this->getPath(), $message, FILE_APPEND);
-            if ($this->isRecording) {
-                $this->playback[] = $message;
-            }
-            if ($this->console) {
-                $this->console->write($message);
+    public function log($message, $label = '')
+    {
+        if ($this->scopeConfig->getValue(TaxjarConfig::TAXJAR_DEBUG, \Magento\Store\Model\ScopeInterface::SCOPE_STORE, $this->storeManager->getStore()->getId())) {
```
fastdivision: This line seems a bit long. Should we assign a variable and / or break those args on to multiple lines?

Model/Logger.php
```diff
@@ -77,24 +110,33 @@ public function getPath()
...
+            try {
+                if (!empty($label)) {
+                    $label = '[' . strtoupper($label) . '] ';
+                }
+                $timestamp = date('d M Y H:i:s', time());
```
fastdivision: Let's continue to clean this method up. Newline above `$timestamp`

Model/Logger.php
```diff
@@ -77,24 +110,33 @@ public function getPath()
...
+                $message = sprintf('%s%s - %s%s', PHP_EOL, $timestamp, $label, $message);
+
+                if (!is_dir(dirname($this->getPath()))) {
+                    // dir doesn't exist, make it
+                    mkdir(dirname($this->getPath()));
+                }
+
+                $this->driverFile->filePutContents($this->getPath(), $message, FILE_APPEND);
```
fastdivision: Newline below this

Model/Smartcalcs.php
```diff
@@ -202,15 +209,27 @@ public function getTaxForOrder(
...
+            $this->logger->log('Calculating sales tax: ' . json_encode($order), 'post');
 
             $this->_setSessionData('order', json_encode($order));
 
             try {
                 $response = $client->request('POST');
                 $this->response = $response;
                 $this->_setSessionData('response', $response);
+
+                if ($response->getStatus() ==200 ) // Success
```
fastdivision: Space between `==` and `200`, braces on same line for conditionals similar to M1

Model/Transaction.php
```diff
@@ -18,6 +18,7 @@
 namespace Taxjar\SalesTax\Model;
 
 use Taxjar\SalesTax\Model\Configuration as TaxjarConfig;
+use Taxjar\SalesTax\Model\Tax\Sales\Total\Quote\Tax;
```
fastdivision: Do we need to include this line for logging?

PR #36, opened 2018-10-02
Fix Partial Refund Transaction Sync

fastdivision:
This PR better supports credit memos with partial refunds, either modified quantities by line item or fixed adjustments applied to the order. Shipping discounts are unavailable for credit memos, so the remainder of the total discount after subtracting line item discounts is used instead.


fastdivision:
PR tested and appears to be working correctly 👍

PR #37, opened 2018-10-15
Update the requireJS mapping by appending taxjar to avoid namespace collissions

ryreeves:
Rename "clipboard" to "taxjarClipboard" and "popup" to "taxjarPopup" in requirejs-config.js


fastdivision:
Tested locally, SmartCalcs Connect popup and transaction sync clipboard copy is working 👍

PR #43, opened 2018-11-20
When rendering orders/credit memos in the admin grids, ensure the orders exist

ryanreeves-taxjar:
In order to ensure the admin grids load reliably we need to confirm the orders exist.   We do this be catching a NoSuchEntityException when loading the orders and credit memos by ID.

### DIFF
--- a/Ui/Component/Listing/Column/SyncedCreditmemo.php
+++ b/Ui/Component/Listing/Column/SyncedCreditmemo.php
@@ -7,6 +7,7 @@ use \Magento\Framework\View\Element\UiComponentFactory;
 use \Magento\Ui\Component\Listing\Columns\Column;
 use \Magento\Framework\Api\SearchCriteriaBuilder;
 use \Magento\Framework\Stdlib\DateTime\Timezone;
+use \Magento\Framework\Exception\NoSuchEntityException;
 
 class SyncedCreditmemo extends Column
 {
@@ -57,8 +58,13 @@ class SyncedCreditmemo extends Column
     {
         if (isset($dataSource['data']['items'])) {
             foreach ($dataSource['data']['items'] as & $item) {
-                $creditmemo = $this->creditmemoRepository->get($item['entity_id']);
                 $creditmemoSyncDate = '';
+
+                try {
+                    $creditmemo = $this->creditmemoRepository->get($item['entity_id']);
+                } catch( NoSuchEntityException $e) {
+                    $this->logger->log($e->getMessage() . ', entity id: ' . $item['entity_id']);
+                }
 
                 if ($creditmemo->getTjSalestaxSyncDate()) {
                     $creditmemoSyncDate = $this->timezone->formatDate(
--- a/Ui/Component/Listing/Column/SyncedCreditmemo.php
+++ b/Ui/Component/Listing/Column/SyncedCreditmemo.php
@@ -39,6 +39,7 @@ class SyncedCreditmemo extends Column
      * @param CreditmemoRepositoryInterface $creditmemoRepository
      * @param SearchCriteriaBuilder $criteria
      * @param Timezone $timezone
+     * @param \Taxjar\SalesTax\Model\Logger $logger
      * @param array $components
      * @param array $data
      */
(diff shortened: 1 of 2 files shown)


Ui/Component/Listing/Column/SyncedCreditmemo.php
```diff
@@ -57,9 +58,14 @@ public function prepareDataSource(array $dataSource)
     {
         if (isset($dataSource['data']['items'])) {
             foreach ($dataSource['data']['items'] as & $item) {
-                $creditmemo = $this->creditmemoRepository->get($item['entity_id']);
                 $creditmemoSyncDate = '';
 
+                try {
+                    $creditmemo = $this->creditmemoRepository->get($item['entity_id']);
+                } catch( NoSuchEntityException $e) {
```
fastdivision: Space between `catch` and `(`

Ui/Component/Listing/Column/SyncedCreditmemo.php
```diff
@@ -7,6 +7,7 @@
 use \Magento\Ui\Component\Listing\Columns\Column;
 use \Magento\Framework\Api\SearchCriteriaBuilder;
 use \Magento\Framework\Stdlib\DateTime\Timezone;
+use \Magento\Framework\Exception\NoSuchEntityException;
 
 class SyncedCreditmemo extends Column
```
fastdivision: Do we need to set the logger filename in this constructor as well?

Ui/Component/Listing/Column/SyncedOrder.php
```diff
@@ -57,9 +68,14 @@ public function prepareDataSource(array $dataSource)
...
-                $order = $this->orderRepository->get($item['entity_id']);
                 $orderSyncDate = '';
 
+                try {
+                    $order = $this->orderRepository->get($item['entity_id']);
+                } catch (NoSuchEntityException $e) {
+                    $this->logger->log($e->getMessage() . ', entity id: ' . $item['entity_id']);
+                }
+
                 if ($order->getTjSalestaxSyncDate()) {
```
fastdivision: If the exception is caught, `$order` may not be set. Should we check to see if `$order` exists for this conditional to avoid an additional error?

```php
if ($order && $order->getTjSalestaxSyncDate())
```

Ui/Component/Listing/Column/SyncedCreditmemo.php
```diff
@@ -57,9 +58,14 @@ public function prepareDataSource(array $dataSource)
...
-                $creditmemo = $this->creditmemoRepository->get($item['entity_id']);
                 $creditmemoSyncDate = '';
 
+                try {
+                    $creditmemo = $this->creditmemoRepository->get($item['entity_id']);
+                } catch( NoSuchEntityException $e) {
+                    $this->logger->log($e->getMessage() . ', entity id: ' . $item['entity_id']);
+                }
+
                 if ($creditmemo->getTjSalestaxSyncDate()) {
```
fastdivision: If the exception is caught, `$creditmemo` may not be set. Should we check to see if `$creditmemo` exists for this conditional to avoid an additional error?

```php
if ($creditmemo && $creditmemo->getTjSalestaxSyncDate())
```

Ui/Component/Listing/Column/SyncedCreditmemo.php
ryanreeves-taxjar: Yes we need the same includes/constructor statements used in SyncedOrder.

Ui/Component/Listing/Column/SyncedCreditmemo.php
```diff
@@ -25,6 +28,11 @@ class SyncedCreditmemo extends Column
...
     protected $timezone;
 
+    /**
+     * @var \Taxjar\SalesTax\Model\Logger
+     */
+    protected $logger;
+
     /**
      * @param ContextInterface $context
      * @param UiComponentFactory $uiComponentFactory
```
fastdivision: Missing `* @param \Taxjar\SalesTax\Model\Logger $logger` in docblock

fastdivision:
LGTM 👍

PR #50, opened 2019-01-10
Fix notice when iterating over empty line items

ryanreeves-taxjar:
Add a check to ensure the line items array is initialized before we
attempt to iterate over it to prevent a notice from being thrown.


Model/Transaction/Refund.php
```diff
@@ -74,8 +74,10 @@ public function build(
             $this->buildLineItems($order, $creditmemo->getAllItems(), 'refund')
         );
 
-        foreach ($this->request['line_items'] as $lineItem) {
-            $itemDiscounts += $lineItem['discount'];
+        if(isset($this->request['line_items'])) {
```
fastdivision: Space between `if` and `(`

PR #52, opened 2019-01-10
Make Nexus street, city and postcode optional

ryanreeves-taxjar:
Change Nexus Address street, city and postcode to be optional and
disable the warning when they are missing.


fastdivision:
Click-tested this PR. Nexus CRUD operations worked properly. Nexuses passed to our API for calculations successfully. Notice with missing zip codes no longer displayed.

PR #53, opened 2019-01-17
Ensure country/state is set when syncing TJ nexus

ryanreeves-taxjar:
When syncing nexus addresses from TaxJar, ensure the country is set
before saving the address. If the country is US/Canada, ensure the
state is also set.


fastdivision:
State validation works properly for US / Canada on nexus address form. When syncing nexus addresses from TaxJar to Magento, the extension no longer creates a nexus with only a zip code or returns an error message with an incomplete business profile in TaxJar 👍

PR #57, opened 2019-02-13
Create var/tmp if the directory doesn't exist

ryanreeves-taxjar:
Before writing to the var/tmp directory, check to see if it exists.  If
it does not, create it.


Observer/ImportRates.php
```diff
@@ -241,6 +241,11 @@ private function _importRates()
         $this->_purgeRates();
 
         try {
+            $dir = $this->directoryList->getPath(DirectoryList::TMP);
```
fastdivision: Add newline below between assignment and conditional

PR #58, opened 2019-02-19
Product tax classes link causes 404 error

ryanreeves-taxjar:
The Product Tax Classes link (admin > Stores > Configuration > Sales >
Tax) pointed to an incorrect URL and caused a 404 error.


fastdivision:
Tested locally, link works now.

PR #63, opened 2019-06-06
Enforce rounding prices and taxes

ryanreeves-taxjar:
Tier prices have the option to apply a percent discount to an items
price which isn't limited by the typical decimal precision.  To ensure
there are no rounding errors, force rounding all prices and tax
calculations based on the default precision.


fastdivision (changes requested):
Waiting on specs

PR #71, opened 2019-07-22
Remove 99999s when product tax class is "none"

ryanreeves-taxjar:
Previously products with a tax class of "none" were assigned 99999 to
mark them as tax exempt.  This change removes that functionality and
requires users to pick a valid tax class to mark the product as exempt.
The "none" tax class now defaults to fully taxable.

### DIFF
--- a/etc/adminhtml/di.xml
+++ b/etc/adminhtml/di.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
-    <virtualType name="Magento\Catalog\Ui\DataProvider\Product\Form\Modifier\Pool">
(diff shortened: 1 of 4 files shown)


etc/adminhtml/di.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
+    <virtualType name="Magento\Catalog\Ui\DataProvider\Product\Form\Modifier\Pool">
```
fastdivision: @ryanreeves-taxjar Do we need a `type` attribute on this tag as well?

ryanreeves-taxjar: Core uses this virtualType with and without the "type" attribute, but I don't see any reason not to add it.

fastdivision (approved):
LGTM 👍

PR #72, opened 2019-07-24
Ensure address validation button is visible (admin)

ryanreeves-taxjar:
Address validation was accidentally firing when searching for customers
during the order creation process.  Ensuring the button is visible
ensures it was actually clicked.

### DIFF
--- a/view/adminhtml/web/js/suggested_addresses_admin.js
+++ b/view/adminhtml/web/js/suggested_addresses_admin.js
@@ -118,16 +118,13 @@ define([
 
         appendButtonToOrder: function () {
             var self = this;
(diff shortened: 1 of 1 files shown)


view/adminhtml/web/js/suggested_addresses_admin.js
```diff
@@ -124,7 +124,10 @@ define([
 
             button.unbind('click').bind('click', function (e) {
                 e.preventDefault();
-                self.validateAddress($(this));
+
+                if (!button.is(':hidden')) {
```
fastdivision: @ryanreeves-taxjar Can you opt for `button.is(':visible')` instead so we can drop the `!` and make it easier to read? Thank you!

fastdivision (approved):
Much better solution 👍

PR #74, opened 2019-08-07
AfterFiles plugin breaks when deploying production

ryanreeves-taxjar:
When setting the deployment mode to production, the call in AfterFiles
to $this->state->getAreaCode() always threw an exception because the
areacode is never set with CLI commands.  Getting the areacode from
$theme works around this issue.

### DIFF
--- a/Block/Adminhtml/AddressValidation.php
+++ b/Block/Adminhtml/AddressValidation.php
@@ -69,6 +69,7 @@ class AddressValidation extends Field
         if (!$this->isAuthorized()) {
             $element->setDisabled('disabled');
         }
+
         $this->_cacheElementValue($element);
 
         return parent::_getElementHtml($element) . $this->_toHtml();
--- a/Block/Adminhtml/AddressValidation.php
+++ b/Block/Adminhtml/AddressValidation.php
@@ -19,12 +19,18 @@ namespace Taxjar\SalesTax\Block\Adminhtml;
 
 use Magento\Backend\Block\Template\Context;
 use Magento\Config\Block\System\Config\Form\Field;
+use Magento\Framework\App\CacheInterface;
 use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\Data\Form\Element\AbstractElement;
 use Taxjar\SalesTax\Model\Configuration as TaxjarConfig;
 
 class AddressValidation extends Field
 {
+    /**
+     * @var \Magento\Framework\App\CacheInterface
+     */
+    protected $cache;
+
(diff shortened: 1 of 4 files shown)


Observer/ConfigReview.php
```diff
@@ -117,4 +118,20 @@ private function _reviewNexusAddresses()
...
+     * @SuppressWarnings(Generic.Files.LineLength.TooLong)
+     */
+    private function _reviewAddressValidation()
+    {
+        $enabled = $this->scopeConfig->getValue(TaxjarConfig::TAXJAR_ADDRESS_VALIDATION);
+        $prevEnabled = $this->cache->load('taxjar_salestax_config_address_validation');
+
+        if (isset($prevEnabled) && $prevEnabled != $enabled) {
+            // @codingStandardsIgnoreStart
+            $this->messageManager->addErrorMessage(__('Please redeploy production mode (bin/magento deploy:mode:set production) to ensure Address Validation works correctly.'));
```
fastdivision: @ryanreeves-taxjar How about this?

> Please redeploy production mode in the Magento CLI (`$ bin/magento deploy:mode:set production`) to ensure address validation works correctly.

Block/Adminhtml/AddressValidation.php
```diff
@@ -60,10 +69,23 @@ protected function _getElementHtml(AbstractElement $element)
         if (!$this->isAuthorized()) {
             $element->setDisabled('disabled');
         }
+        $this->_cacheElementValue($element);
```
fastdivision: Newline above 72

Observer/ConfigReview.php
```diff
@@ -117,4 +118,20 @@ private function _reviewNexusAddresses()
...
+     * @SuppressWarnings(Generic.Files.LineLength.TooLong)
+     */
+    private function _reviewAddressValidation()
+    {
+        $enabled = $this->scopeConfig->getValue(TaxjarConfig::TAXJAR_ADDRESS_VALIDATION);
+        $prevEnabled = $this->cache->load('taxjar_salestax_config_address_validation');
+
+        if (isset($prevEnabled) && $prevEnabled != $enabled) {
+            // @codingStandardsIgnoreStart
+            $this->messageManager->addErrorMessage(__('Please redeploy production mode in the Magento CLI ($ bin/magento deploy:mode:set production) to ensure address validation works correctly.'));
```
fastdivision: @ryanreeves-taxjar Just tested this locally, I would go with a warning message via `addWarningMessage` instead of an error message here.

PR #75, opened 2019-08-13
Pi 244 replace serialize functions

ryanreeves-taxjar:
Replace direct calls to serialize and uses of deprecated Unserialize
class with the Serialize class

### DIFF
--- a/Setup/UpgradeData.php
+++ b/Setup/UpgradeData.php
@@ -23,21 +23,55 @@ use Magento\Eav\Setup\EavSetup;
 use Magento\Framework\Setup\UpgradeDataInterface;
 use Magento\Framework\Setup\ModuleDataSetupInterface;
 use Magento\Framework\Setup\ModuleContextInterface;
+use Taxjar\SalesTax\Model\ClientFactory;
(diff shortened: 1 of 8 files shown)


Setup/UpgradeData.php
```diff
@@ -163,5 +197,18 @@ public function upgrade(
             $lastSyncCodeId->setData('used_in_forms', ['adminhtml_customer',]);
             $lastSyncCodeId->getResource()->save($lastSyncCodeId);
         }
+
+        if (version_compare($context->getVersion(), '1.0.3', '<')) {
+            $configJson = $this->client->getResource('config');
```
fastdivision: @ryanreeves-taxjar Do we need to check for an API token before hitting this endpoint?

Setup/UpgradeData.php
```diff
@@ -163,5 +197,18 @@ public function upgrade(
...
+
+            if (isset($configJson['configuration']) && isset($configJson['configuration']['states'])) {
```
fastdivision: This conditional differs from M1, should it be the same?
```php
if (is_array($configJson) && isset($configJson['configuration']) && isset($configJson['configuration']['states'])) {
```

PR #76, opened 2019-08-29
Change to namespace for phpunit 6.x compatibility

uabassguy:


ryanreeves-taxjar:
Thanks @uabassguy!  We'll take a look into this and let you know what we find.

PR #78, opened 2019-09-04
Fail "hasNexus" check if region is empty

ryanreeves-taxjar:
If no region code is provided, the nexus check was returning true even
though nexus may not exist.  Changed it to return false if no region is
provided.


fastdivision (approved):
LGTM 👍

PR #79, opened 2019-09-11
Update error handling when missing address data

ryanreeves-taxjar:
Remove the unused generic error message and return an empty array.  Add
a condition for empty suggested address arrays before attempting to
render them.  Minor code cleanup as well.


fastdivision (approved):
Tested locally on frontend (guest, saved addresses) and backend (new order, customer addresses). Everything looks good to me 👍

PR #82, opened 2019-09-19
Include checking for B2B editions as well

ryanreeves-taxjar:
Support gift cards that come from Magento's B2B mode as well as
Enterprise editions.


fastdivision (approved):
LGTM 👍

PR #88, opened 2019-10-29
Update customer observer with customer repository

ryanreeves-taxjar:
Loading customer data directly from the customer model led to situations
where incorrect address data was used.  Using the repository ensures the
correct address data is loaded.


Observer/Customer.php
```diff
@@ -56,23 +61,26 @@ class Customer implements ObserverInterface
...
     /**
+     * @param \Magento\Customer\Api\AddressRepositoryInterface $addressRepository
      * @param \Taxjar\SalesTax\Model\ClientFactory $clientFactory
-     * @param \Magento\Customer\Model\CustomerFactory $customerFactory
+     * @param \Magento\Customer\Api\CustomerRepositoryInterface $customerRepository
      * @param \\Magento\Framework\Stdlib\DateTime\DateTime $date
      * @param \Magento\Directory\Model\RegionFactory $regionFactory
      *
      */
     public function __construct(
```
dallendalton: Constructor param documentation needs updated to match actual params.

PR #94, opened 2019-11-28
Don't recalculate tax on quote items when module is disabled

Maikel-Koek:
When creating multiple orders in the same process, the tax was recalculated for quotes which were in a store where TaxJar was disabled, because even though the `$isEnabled` check was used for getting tax, the `isValidResponse` succeeded, so the quote's tax was still recalculated.

Now when the module's disabled, just skip all and return to parent immediately.

### DIFF
--- a/Model/Tax/Sales/Total/Quote/Tax.php
+++ b/Model/Tax/Sales/Total/Quote/Tax.php
@@ -119,10 +119,12 @@ class Tax extends \Magento\Tax\Model\Sales\Total\Quote\Tax
             $quote->getStoreId()
         );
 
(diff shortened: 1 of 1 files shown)


ryanreeves-taxjar:
Thanks for this contribution @Maikel-Koek!  We're going to review it internally and let you know what we find!

NickdeK:
Any news on this @ryanreeves-taxjar?

ryanreeves-taxjar:
@Maikel-Koek @NickdeK We really appreciate your patience on this.  We're going to review it internally a bit more but I don't foresee any issues with this PR.

PR #98, opened 2020-01-17
PTC Expansion Refactor

ryanreeves-taxjar:
Storing product tax categories as a configuration value won't scale well
as we continue to increase the total number of categories. Reworking
categories to be stored in their own table eliminates this problem and
will support significantly more categories.

### DIFF
--- a/Setup/UpgradeData.php
+++ b/Setup/UpgradeData.php
@@ -20,24 +20,24 @@ namespace Taxjar\SalesTax\Setup;
 use Magento\Customer\Api\CustomerMetadataInterface;
 use Magento\Eav\Model\Config;
 use Magento\Eav\Setup\EavSetup;
+use Magento\Framework\Setup\ModuleContextInterface;
+use Magento\Framework\Setup\ModuleDataSetupInterface;
 use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\Setup\UpgradeDataInterface;
-use Magento\Framework\Setup\ModuleDataSetupInterface;
-use Magento\Framework\Setup\ModuleContextInterface;
 use Taxjar\SalesTax\Model\ClientFactory;
 use Taxjar\SalesTax\Model\Configuration as TaxjarConfig;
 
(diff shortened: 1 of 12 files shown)


Setup/UpgradeData.php
```diff
@@ -221,5 +229,10 @@ public function upgrade(
                 }
             }
         }
+
+        if (version_compare($context->getVersion(), '1.0.4', '<')) {
```
codymullins: I would add a comment here to explain why this happens for this version.

codymullins: In the UpgradeScheme file, too.

Setup/UpgradeSchema.php
```diff
@@ -86,5 +86,54 @@ public function upgrade(SchemaSetupInterface $setup, ModuleContextInterface $con
 
             $installer->endSetup();
         }
+
+        if (version_compare($context->getVersion(), '1.0.3', '<')) {
```
codymullins: Is this method idempotent?

Setup/UpgradeData.php
ryanreeves-taxjar: @codymullins We don't have comments for any the upgrade steps.  Is there a reason you think this version deserves one?  Or are you suggesting each version should have one?

Setup/UpgradeSchema.php
ryanreeves-taxjar: @codymullins  The code inside `version_compare` will only run once, with the module's version being updated after the upgrade script finishes.  It isn't technically idempotent, as it makes an API call which can have different results over time.

Setup/UpgradeData.php
codymullins: Yeah, it might be good in general - didn't mean "this version especially." For others diagnosing issues and looking through that aren't familiar with the code/history.

Setup/UpgradeSchema.php
codymullins: Understood. Is this the common practice with Magento?

PR #99, opened 2020-01-23
Fix division by zero when issuing credit memos

ryanreeves-taxjar:
When issuing a credit memo for an unshipped order, credit memos weren't
correctly calculating quantity or amount refunded.  Loading the order
item from the order returns the correct item data.  Also adds a check
to skip the item if quantity is actually zero.

### DIFF
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -238,8 +238,13 @@ class Transaction
             ];
 
             if ($type == 'refund' && method_exists($item, 'getOrderItem')) {
-                $orderItem = $item->getOrderItem();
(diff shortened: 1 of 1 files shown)


Model/Transaction.php
```diff
@@ -238,8 +238,13 @@ protected function buildLineItems($order, $items, $type = 'order') {
             ];
 
             if ($type == 'refund' && method_exists($item, 'getOrderItem')) {
```
mdrobbins: Should this be checking for the existence of 'getOrderItemId' now?

fastdivision: @ryanreeves-taxjar Checking this out

PR #100, opened 2020-01-28
Improve support for M2E Pro

ryanreeves-taxjar:
When syncing orders/refunds to TaxJar, first check to see if M2ePro is
installed.  If so, try to load the original provider from the order.

### DIFF
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -90,7 +90,6 @@ class Transaction
 
         $this->client = $this->clientFactory->create();
         $this->client->showResponseErrors(true);
-
     }
 
     /**
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -58,12 +58,18 @@ class Transaction
     protected $client;
 
     /**
+     * @var \Magento\Framework\ObjectManagerInterface
+     */
+    protected $objectManager;
+
+    /**
      * @param \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig
      * @param \Taxjar\SalesTax\Model\ClientFactory $clientFactory
      * @param \Magento\Catalog\Model\ProductRepository $productRepository
      * @param \Magento\Directory\Model\RegionFactory $regionFactory
      * @param \Magento\Tax\Api\TaxClassRepositoryInterface $taxClassRepository
      * @param \Taxjar\SalesTax\Model\Logger $logger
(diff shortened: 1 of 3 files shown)


Model/Transaction.php
```diff
@@ -274,6 +283,32 @@ function buildCustomerExemption($order)
...
+
+        try {
+            if (class_exists('\Ess\M2ePro\Model\Order')) {
+                $m2eOrder = $this->objectManager->create('\Ess\M2ePro\Model\Order');
+                $m2eOrder = $m2eOrder->load($order->getId(), 'magento_order_id');
+            }
+
+            if (isset($m2eOrder)) {
+                if (in_array($m2eOrder->getMarketplace()->getComponentMode(), ['amazon', 'ebay', 'walmart'])) {
+                    $provider = $m2eOrder->getMarketplace()->getComponentMode();
```
ryanreeves-taxjar: Confirm getComponentMode() loads correctly.

Model/Transaction.php
```diff
@@ -57,28 +57,36 @@ class Transaction
...
      */
     public function __construct(
         \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
         \Taxjar\SalesTax\Model\ClientFactory $clientFactory,
         \Magento\Catalog\Model\ProductRepository $productRepository,
         \Magento\Directory\Model\RegionFactory $regionFactory,
         \Magento\Tax\Api\TaxClassRepositoryInterface $taxClassRepository,
-        \Taxjar\SalesTax\Model\Logger $logger
+        \Taxjar\SalesTax\Model\Logger $logger,
+        \Magento\Framework\ObjectManagerInterface $objectmanager
```
fastdivision: @ryanreeves-taxjar Consider camel casing `$objectmanager` to `$objectManager`

Model/Transaction.php
```diff
@@ -274,6 +282,30 @@ function buildCustomerExemption($order)
         return [];
     }
 
+    /**
+     * @param \Magento\Sales\Model\Order $order
+     * @return string
+     */
+    public function getProvider($order)
```
fastdivision: @ryanreeves-taxjar This function should have a `protected` scope.

ryanreeves-taxjar: I made this public when we were first looking at using a plugin around this function and forgot to change it back.

fastdivision:
Tested locally, no issues passing default provider ID with M2E disabled 👍

PR #101, opened 2020-01-30
Changing customer save observer region code getter

andrew-ebsco:
$customerAddress->getRegion()->getRegionCode() stopped working in Magento 2.3.4. Using $customerAddress->getRegionCode() instead.

### DIFF
--- a/Observer/Customer/Save.php
+++ b/Observer/Customer/Save.php
@@ -63,7 +63,7 @@ class Save extends Customer
         if ($customerAddress) {
             $data = array_merge($data, [
                 'country' => $customerAddress->getCountryId(),
-                'state' => $customerAddress->getRegion()->getRegionCode(),
+                'state' => $customerAddress->getRegionCode(),
                 'zip' => $customerAddress->getPostcode(),
                 'city' => $customerAddress->getCity(),
                 'street' => implode(", ", $customerAddress->getStreet())


fastdivision:
Thanks @andrew-ebsco! We'll check this out and get it merged shortly.

andrew-ebsco:
Hey @ryanreeves-taxjar, the bug appeared for us when submitting an admin order for a new customer. Saving the new customer in the order submission triggers the observer.

The Address model extends from an [abstract class](https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Customer/Model/Address/AbstractAddress.php#L390). In the abstract class, getRegion is defined to return a string not an object so I switched to using getRegionCode instead.

srsathish92:
@ryanreeves-taxjar, We are also facing the same issue when placing the order from Admin with Magento 2.3.3.

**Error message:** `Error: Uncaught exception 'Error' with message 'Call to a member function getRegionCode() on string' in /vendor/taxjar/module-taxjar/Observer/Customer/Save.php:66`

ryanreeves-taxjar:
Thanks @andrew-ebsco  and @srsathish92.  

It looks like the issue stems from the address returned from the AddressRepository.  When creating a new customer or editing a customer **without a default shipping address**, `$customer->getAddresses()` returns instances of `\Magento\Customer\Model\Address`.  When editing an existing customer **with a default shipping address**, the AddressRepository returns an instance of `\Magento\Customer\Model\Data\Address`.   

@andrew-ebsco Unfortunately the fix you've provided breaks when editing a customer with a saved default shipping address.  I'm working on a fix that works in both instances and should have a PR opened for it very soon.  

Thanks to both of you for reporting this issue and for working with us to investigate it in more detail!

srsathish92:
@ryanreeves-taxjar Thanks for the confirmation. when can we expect the patch?

ryanreeves-taxjar:
New PR is available here:  https://github.com/taxjar/taxjar-magento2-extension/pull/105

ryanreeves-taxjar:
Thanks @andrew-ebsco for your PR.  We've addressed the issue in https://github.com/taxjar/taxjar-magento2-extension/pull/105.  If you notice any issues please let us know!

PR #102, opened 2020-01-31
Fix incorrect qty with multiple credit memos

ryanreeves-taxjar:
When issuing  multiple refunds, the $lineItem['qty'] was the total of all refunds.
This caused the qty on the first refund to be correct, and each subsequent refund
was inaccurate. Getting  the qty from the credit memo, instead of the order, fixes
this issue.


fastdivision:
@ryanreeves-taxjar Did we verify the other line item attributes return the correct values across multiple credit memos such as `$orderItem->getAmountRefunded()` for `unit_price`?

PR #105, opened 2020-02-04
Fix customer save observer without default address

ryanreeves-taxjar:
### Context

When saving a customer, the observer uses different address classes depending on if it's the default shipping address or not.  This change ensures we properly load the region from either class. 

### Description

Because the function `getRegion` returns a string in `Magento\Customer\Model\Address`, it caused a fatal error when trying to load the regionCode from it.  

### Performance

This change adds a single if/ifelse statement that should have no noticeable impact on performance.

### Testing

There are two testing paths that should both be tested:

1. Create a new order in the admin and create a new customer (instead of using an existing one).  After entering the address information, ensure the "save customer address" box is checked.  After completing the order, ensure no error is thrown and the region (US state) correctly syncs to TaxJar.  

Screen Shot 2020-02-04 at 12 24 29 PM

2.  Edit an existing customer with a default shipping address set.  Save the customer and ensure no error is thrown and the region correctly syncs to TaxJar.  

Screen Shot 2020-02-04 at 12 31 03 PM

Screen Shot 2020-02-04 at 12 31 31 PM

#### Versions

- [X] Magento 2.4
- [X] Magento 2.3
- [X] Magento 2.2
- [X] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Observer/Customer/Save.php
+++ b/Observer/Customer/Save.php
@@ -63,11 +63,17 @@ class Save extends Customer
         if ($customerAddress) {
             $data = array_merge($data, [
                 'country' => $customerAddress->getCountryId(),
-                'state' => $customerAddress->getRegion()->getRegionCode(),
+                'state' => '',
                 'zip' => $customerAddress->getPostcode(),
                 'city' => $customerAddress->getCity(),
                 'street' => implode(", ", $customerAddress->getStreet())
             ]);
+
+            if (get_class($customerAddress) == 'Magento\Customer\Model\Address') {
+                $data['state'] = $customerAddress->getRegionCode();
+            } elseif (get_class($customerAddress) == 'Magento\Customer\Model\Data\Address') {
+                $data['state'] = $customerAddress->getRegion()->getRegionCode();
+            }
         }
 
         $response = $this->updateTaxjar($customer->getTjLastSync(), $data);
--- a/Observer/Customer/Save.php
+++ b/Observer/Customer/Save.php
@@ -63,7 +63,6 @@ class Save extends Customer
         if ($customerAddress) {
             $data = array_merge($data, [
                 'country' => $customerAddress->getCountryId(),
-                'state' => '',
                 'zip' => $customerAddress->getPostcode(),
                 'city' => $customerAddress->getCity(),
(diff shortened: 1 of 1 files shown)


Observer/Customer/Save.php
```diff
@@ -63,11 +63,17 @@ public function execute(Observer $observer)
         if ($customerAddress) {
             $data = array_merge($data, [
                 'country' => $customerAddress->getCountryId(),
-                'state' => $customerAddress->getRegion()->getRegionCode(),
+                'state' => '',
```
codymullins: Would it be better to leave this out or use `null`?

Observer/Customer/Save.php
```diff
@@ -63,11 +63,17 @@ public function execute(Observer $observer)
...
                 'zip' => $customerAddress->getPostcode(),
                 'city' => $customerAddress->getCity(),
                 'street' => implode(", ", $customerAddress->getStreet())
             ]);
+
+            if (get_class($customerAddress) == 'Magento\Customer\Model\Address') {
+                $data['state'] = $customerAddress->getRegionCode();
+            } elseif (get_class($customerAddress) == 'Magento\Customer\Model\Data\Address') {
+                $data['state'] = $customerAddress->getRegion()->getRegionCode();
+            }
```
codymullins: Are there any conditions here that would fall into an `else` we should be checking? Would this be an error state or require us to throw an exception to ensure we don't have a missing state field?

Observer/Customer/Save.php
ryanreeves-taxjar: @codymullins Good catch.  We set $data['state'] to null initially, so there's no reason to redeclare it here.

Observer/Customer/Save.php
ryanreeves-taxjar: @codymullins There are additional conditions I can think of (outside of a 3rd party extension that extends the Address model).  Ideally Magento's `AbstractAddress` and api data `AddressInterface` classes would be compatible (they both utilize the same data) but that isn't the case here.

PR #116, opened 2020-03-03
Sort by sync date broken in order grid

bjwirgau:
### Context
An exception is thrown when sorting by sync date in the order grid. The error thrown is the following:
`SQLSTATE[42S22]: Column not found: 1054 Unknown column 'synced' in 'order clause', query was: SELECT `main_table`.* FROM `sales_order_grid` AS `main_table` ORDER BY synced ASC`

The sync column is missing from the order grid table and does not have any way of sorting by this column.

### Expected Results
When the sync column is visible, the sorting feature should sort by the sync date correctly.

### Description
The fix adds a column to the sales_order_grid table and maps the values from the order table to the order grid table. In order to have the data populated into the grid correctly after sorting, the sales_order_grid table needs to be truncated and regenerated. The table will be regenerated through the cron process `sales_grid_order_async_insert`. This can be updated either through cron or with a magerun command `n98-magerun2 sys:cron:run sales_grid_order_async_insert`. Note, asynchronous indexing needs to be enabled to regenerate the order grid table. The config is under Stores -> Configuration -> Advanced -> Developer -> Grid Settings -> Asynchronous Indexing

### Testing
1. Go to the order table in the admin panel.
2. Make column "Synced to Taxjar" visible in the order grid if it is not visible.
3. Sort by synced to taxjar and verify that the grid correctly sorts the orders.

#### Versions
Taxjar 1.4.9
Magento Open Source 2.3.2
PHP 7.2

### DIFF
--- a/Setup/UpgradeSchema.php
+++ b/Setup/UpgradeSchema.php
@@ -145,7 +145,7 @@ class UpgradeSchema implements UpgradeSchemaInterface
              */
             $installer->getConnection()->addColumn(
                 $installer->getTable('sales_order_grid'),
-                'tj_salestax_sync_date',
+                'synced',
                 [
                     'type' => \Magento\Framework\DB\Ddl\Table::TYPE_TIMESTAMP,
                     'nullable' => true,
--- a/Setup/UpgradeSchema.php
+++ b/Setup/UpgradeSchema.php
@@ -135,5 +135,25 @@ class UpgradeSchema implements UpgradeSchemaInterface
 
             $installer->endSetup();
         }
+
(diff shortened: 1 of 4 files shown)


ryanreeves-taxjar:
Thanks @bjwirgau!  We'll take a look into this and let you know what we find!

ryanreeves-taxjar:
Hi @bjwirgau.  We've come up with an alternative solution to this problem.  Please feel free to take a look at https://github.com/taxjar/taxjar-magento2-extension/pull/135 and leave us any feedback!

PR #117, opened 2020-03-18
Check if transaction sync is disabled before sync

ryanreeves-taxjar:
### Context
This PR limits syncing transactions (aka backfilling) when a specific website/store has disabled transaction sync.  It also restricts the transactions that are backfilled to the current scope config in the admin.

### Description
This PR moves the checks to see if a transaction can be backfilled to the `isSyncable` method.  It checks the store associated with the specific order confirm transaction sync is enabled for that store.  

### Performance
This PR should improve performance as it reduces the number of transactions synced in certain situations.  

### Testing
1. Ensure the "Transaction Sync" admin configuration (Stores > Configuration > Sales > Tax > TaxJar) is set to "Yes"
Screen Shot 2020-03-19 at 1 50 36 PM
2. Change the admin scope to a website or store view.  
Screen Shot 2020-03-19 at 1 48 58 PM
3.  Change "Transaction Sync" from "Yes" to "No"
4. Change the scope back to "Default" and confirm that associated transactions don't sync.  (ex:  #2000000001)
Screen Shot 2020-03-19 at 1 48 52 PM

#### Versions

- [X] Magento 2.3
- [X] Magento 2.2
- [X] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x


Model/Transaction/Backfill.php
```diff
@@ -82,31 +85,30 @@ class Backfill
...
      * @param \Magento\Framework\Api\SearchCriteriaBuilder $searchCriteriaBuilder
      */
     public function __construct(
-        ScopeConfigInterface $scopeConfig,
-        RequestInterface $request,
         TransactionFactory $transactionFactory,
         OrderFactory $orderFactory,
         RefundFactory $refundFactory,
         Logger $logger,
+        \Magento\Backend\Block\Template\Context $context,
```
fastdivision: @ryanreeves-taxjar This dependency was causing a blank config screen and various CLI errors locally. When this dependency is injected, somewhere down the chain it initiates an instance that needs to know the state of the application and triggering the following errors:

```
[Magento\Framework\Exception\SessionException]
Area code not set: Area code must be set before starting a session.

[Magento\Framework\Exception\LocalizedException]
Area code is not set
```

It seems like this should only be injected within a block class. I would restore the original dependencies for `scopeConfig` and `request`. To get `storeManager`, use `Magento\Store\Model\StoreManagerInterface` like elsewhere in the codebase.

ryanreeves-taxjar: m2 3-cli
m2 3-admin
m2 2-cli
m2 1-cli
m2 2-admin
m2 1-admin

@fastdivision I've been unable to reproduce the error in any of my testing.  We've also failed Magento code reviews before for not using context.  

I'll go ahead and back these changes out, but I think this is worth investigating more in the future.

fastdivision: Thanks for investigating @ryanreeves-taxjar. Specific version was `2.2.4`. Let me know if you see any errors after compiling and then running `bin/magento -vvv`.

view/adminhtml/templates/transaction_sync.phtml
```diff
@@ -70,7 +70,7 @@
                 $('#transaction-sync-log').find('> pre').html('Syncing transactions to TaxJar, please wait...');
 
                 $.ajax({
-                    url: '<?php echo $block->escapeUrl($block->getStoreUrl('taxjar/transaction/backfill')) ?>',
+                    url: '<?php echo $block->escapeUrl($block->getStoreUrl('taxjar/transaction/backfill', $block->getRequest()->getParams())) ?>',
```
fastdivision, line 73: @ryanreeves-taxjar On Magento 2.2 this addition returns back the following error in the AJAX response when backfilling from a specific store view:

```
{"error":true,"message":"You entered an invalid Secret Key. Please refresh the page."}
```

I'm seeing `/section/tax/store/7/` appended to the URL before the query params: `/admin/taxjar/transaction/backfill/key/:key/section/tax/store/7/?from_date=04%2F1%2F2020&to_date=&isAjax=true`

Can we pass the params in the `data` object similar to `from_date` and `to_date`?

PR #118, opened 2020-03-31
Max discount amount ignores item quantity

ryanreeves-taxjar:
### Context
Currently if an item's discounts exceed the price of an individual item, the discount is maxed at the price of that item.  This ignores the quantity of the items being ordered and incorrectly calculates the discount.  Ex:

12 items ordered at $10 each with a 10% discount.   The subtotal is $120 with a discount of $12.  Because $12 > $10, the discount will be incorrectly calculated as $10.

### Description
When enforcing the maximum discount amount, the unit price should
also consider the quantity of the items being ordered.

### Performance
This change has a negligible impact on performance.

### Testing
1. Add 11 (or more) taxable items to Magento's cart
2. Apply a 10% (or greater) discount 
3. Ensure that the discount amount is greater than the price of an individual item

#### Versions

- [X] Magento 2.3
- [X] Magento 2.2
- [X] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -200,6 +200,15 @@ class Transaction
             $itemType = $item->getProductType();
             $parentItem = $item->getParentItem();
             $unitPrice = (float) $item->getPrice();
+            $quantity = (int) $item->getQtyOrdered();
+
+            if ($type == 'refund') {
+                $quantity = (int) $item->getQty();
+
+                if ($quantity === 0) {
+                    continue;
+                }
+            }
 
             if (($itemType == 'simple' || $itemType == 'virtual') && $item->getParentItemId()) {
                 if (!empty($parentItem) && $parentItem->getProductType() == 'bundle') {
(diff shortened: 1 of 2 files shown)


Model/Transaction.php
```diff
@@ -200,6 +200,15 @@ protected function buildLineItems($order, $items, $type = 'order') {
             $itemType = $item->getProductType();
             $parentItem = $item->getParentItem();
             $unitPrice = (float) $item->getPrice();
+            $quantity = (int) $item->getQtyOrdered();
+
+            if ($type == 'refund') {
```
ScottRudiger, line 205: Any reason not to use `===` strict equality here?

ryanreeves-taxjar: @ScottRudiger  I think it's ok to use == here because we're doing a string comparison against a variable with a known type (we explicitly set the value of the string).  It's also consistent with how we do string comparisons in the rest of the code.

We do use === when comparing against true, false, and zero, which could accidentally be  coerced to a different type and cause unexpected results.  

That said, I can't think of a compelling reason not to use === everywhere we currently use ==, but I think that would be more appropriate for a different PR.

PR #120, opened 2020-04-02
Fix for credit memo adjustments

ryanreeves-taxjar:
### Context
Adding an adjustment fee or adjustment refund to a credit memo was causing a mismatch between the sum of the line items and the calculated subtotal.  

### Description
This PR distributes the adjustment fee proportionally across the line item discounts, and adds a new line item for the adjustment refund.

### Performance
This change has a negligible impact or performance.  

### Testing
1. Create a new order in Magento
2. Issue a credit memo with an adjustment refund and/or adjustment fee
Screen Shot 2020-04-01 at 6 37 00 PM
3. Ensure that any adjustment fees were distributed across the item discounts and any adjustment refund was added as a new line item. 
Screen Shot 2020-04-02 at 9 43 50 AM

#### Versions

- [X] Magento 2.3
- [X] Magento 2.2
- [X] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Model/Transaction/Refund.php
+++ b/Model/Transaction/Refund.php
@@ -77,8 +77,25 @@ class Refund extends \Taxjar\SalesTax\Model\Transaction
         );
 
         if (isset($this->request['line_items'])) {
-            foreach ($this->request['line_items'] as $lineItem) {
+            $adjustmentFee = $creditmemo->getAdjustmentNegative();
+            $adjustmentRefund = $creditmemo->getAdjustmentPositive();
+
+            foreach ($this->request['line_items'] as $k => $lineItem) {
+                $lineItemSubtotal = $lineItem['unit_price'] * $lineItem['quantity'];
+                $this->request['line_items'][$k]['discount'] += ($adjustmentFee * ($lineItemSubtotal / $creditmemo->getSubtotal()));
                 $itemDiscounts += $lineItem['discount'];
+            }
+
+            if ($adjustmentRefund) {
+                $this->request['line_items'][] = [
+                    'id' => 'adjustment-refund',
+                    'quantity' => 1,
+                    'product_identifier' => 'adjustment-refund',
+                    'description' => 'Adjustment Refund',
+                    'unit_price' => $adjustmentRefund,
+                    'discount' => 0,
+                    'sales_tax' => 0
+                ];
             }
         }
 
--- a/Model/Transaction/Refund.php
+++ b/Model/Transaction/Refund.php
@@ -80,9 +80,10 @@ class Refund extends \Taxjar\SalesTax\Model\Transaction
             $adjustmentFee = $creditmemo->getAdjustmentNegative();
             $adjustmentRefund = $creditmemo->getAdjustmentPositive();
 
+            // Discounts on credit memos act as fees and shouldn't be included in $itemDiscounts
             foreach ($this->request['line_items'] as $k => $lineItem) {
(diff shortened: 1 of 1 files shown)


Model/Transaction/Refund.php
```diff
@@ -77,9 +77,26 @@ public function build(
         );
 
         if (isset($this->request['line_items'])) {
-            foreach ($this->request['line_items'] as $lineItem) {
+            $adjustmentFee = $creditmemo->getAdjustmentNegative();
+            $adjustmentRefund = $creditmemo->getAdjustmentPositive();
+
+            foreach ($this->request['line_items'] as $k => $lineItem) {
+                $lineItemSubtotal = $lineItem['unit_price'] * $lineItem['quantity'];
+                $this->request['line_items'][$k]['discount'] += ($adjustmentFee * ($lineItemSubtotal / $creditmemo->getSubtotal()));
```
fastdivision, line 85: @ryanreeves-taxjar Since we call `$creditmemo->getSubtotal()` at the top of the `build` method we could just use `$subtotal` here right?

Model/Transaction/Refund.php
```diff
@@ -77,9 +77,26 @@ public function build(
...
                 $itemDiscounts += $lineItem['discount'];
```
fastdivision, line 86: @ryanreeves-taxjar You may want to add a comment above this line explaining that we're _not_ including the adjustment discount in `$itemDiscounts`, since the refund-level `$discount` doesn't consider an adjustment a discount and it won't be factored in. At first I thought this could have been a potential issue since we're updating the original line items array right above but not increasing `$itemDiscounts`, but this makes sense.

Model/Transaction/Refund.php
```diff
@@ -77,9 +77,26 @@ public function build(
         );
 
         if (isset($this->request['line_items'])) {
```
fastdivision, line 79: @ryanreeves-taxjar Is it possible for a refund to only have an adjustment with no line items refunded? I imagine in this scenario we'd simply pass the refund-level amounts and this code would be fine as-is. Just checking.

fastdivision (changes requested):

fastdivision:
Waiting on spec in #128

Model/Transaction/Refund.php
ryanreeves-taxjar: Confirmed that this works for credit memos with only adjustments, no items.

Model/Transaction/Refund.php
```diff
@@ -77,9 +77,26 @@ public function build(
...
-            foreach ($this->request['line_items'] as $lineItem) {
+            $adjustmentFee = $creditmemo->getAdjustmentNegative();
```
dallendalton, line 80: Any thoughts on moving this new functionality to its own method?

fastdivision (approved):

ScottRudiger (approved):

Model/Transaction/Refund.php
ryanreeves-taxjar: In a future PR I'd love to refactor a lot of this, once we have more tests in place. 😀

ryanreeves-taxjar:
Changes are live in branch INT-641-integration-testing

ryanreeves-taxjar:
Closing this PR as the code was merged into https://github.com/taxjar/taxjar-magento2-extension/pull/128

PR #139, opened 2020-06-17
Add a sandbox toggle to the admin configuration

ryanreeves-taxjar:
### Description

This change adds a toggle to the Tax admin configuration page, nested under debug mode.  When enabled, all API requests will be made to TaxJar's sandbox.  It also cleans up a small bug in Smartcalcs to standardize the way API calls are made.

Screen Shot 2020-06-18 at 6 44 49 PM

### Performance

Sandbox mode should only be used for testing.  This change has a negligible impact on production environments.  

### Testing

1.  Install the TaxJar extension,  connect the account to TaxJar and follow all setup instructions
2.  Enable debug mode and then enable sandbox mode
3.  Ensure all API requests are made to the sandbox endpoint (api.taxjar.sandbox.com)

#### Versions

- [X] Magento 2.3
- [X] Magento 2.2
- [X] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x


dallendalton (dismissed):
Have we thought about creating a separate field for the sandbox API token as well? This would allow a user to switch between sandbox and prod without having to edit another field. I don't really think it's strictly necessary, just something to think about.

ryanreeves-taxjar:
@dallendalton That's a good idea.  Right now, disabling the sandbox leaves the user without a valid API token.

Model/Smartcalcs.php
```diff
@@ -215,20 +214,12 @@ public function getTaxForOrder(
 
         if ($this->_orderChanged($order)) {
             $client = $this->clientFactory->create();
-            $client->setUri(TaxjarConfig::TAXJAR_API_URL . '/magento/taxes');
-            $client->setConfig([
-                'useragent' => $this->tjHelper->getUserAgent(),
-                'referer' => $this->tjHelper->getStoreUrl()
-            ]);
-            $client->setHeaders('Authorization', 'Bearer ' . $apiKey);
```
fastdivision, line 223: @ryanreeves-taxjar We use `ZendClientFactory` directly here to pass an API token based on the store ID of the quote object:

```php
$apiKey = preg_replace('/\s+/', '', $this->scopeConfig->getValue(TaxjarConfig::TAXJAR_APIKEY,
    \Magento\Store\Model\ScopeInterface::SCOPE_STORE,
    $quote->getStoreId()
));
```

In our client class we get the API token like so:

```php
$this->apiKey = trim($this->scopeConfig->getValue(TaxjarConfig::TAXJAR_APIKEY));
```

Do you anticipate any issues here dropping this code?

ryanreeves-taxjar: @fastdivision I ended up reverting those changes because of some additional issues I encountered during further testing.

ryanreeves-taxjar:
@dallendalton Newest commit adds a separate field for the sandbox API token and reworks loading the token based on the sandbox configuration setting.  LMK if you have any more questions!

etc/adminhtml/system.xml
```diff
@@ -73,6 +73,24 @@
                         <field id="debug">1</field>
                     </depends>
                 </field>
+                <field id="sandbox" translate="label" sortOrder="12" type="select" showInDefault="1" showInWebsite="1" showInStore="1">
+                    <label>Enable Sandbox</label>
+                    <comment>Change all API requests to be made to TaxJar's sandbox account.</comment>
```
fastdivision, line 78: @ryanreeves-taxjar Since our sandbox environment primarily supports API calculations at the moment, should we consider saying something like:

```
Use the TaxJar sandbox environment to test API sales tax calculations.
```

Moreover, what if address validation is enabled? We don't support AV in the sandbox. For transaction sync we'll return stubbed responses.

- Do we need to show a notice / notification in admin when sandbox is enabled?
- Do we need to disable or lock specific functionality such as address validation or syncing nexus addresses when sandbox is enabled?
- Do we need to explain any limitations around the sandbox environment elsewhere or link to our API documentation?

codymullins: @fastdivision I've had several customers confused by your third bullet point as well. I definitely think we need a better strategy here, for at least the messaging aspect.

codymullins (dismissed):

etc/adminhtml/system.xml
ScottRudiger: For address validation, specifically, could we just use their regular API token (if it's still sitting in the other field) even when Sandbox is enabled? That way, there's no visible problem to the customer, unless they don't have a production token in place.

etc/adminhtml/system.xml
```diff
@@ -73,6 +73,24 @@
                         <field id="debug">1</field>
                     </depends>
                 </field>
+                <field id="sandbox" translate="label" sortOrder="12" type="select" showInDefault="1" showInWebsite="1" showInStore="1">
```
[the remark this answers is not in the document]
ryanreeves-taxjar: When the sortOrder is tied, they will render in the order they are defined.  I've updated the sortOrder to be specific.

etc/adminhtml/system.xml
ryanreeves-taxjar: After some review and internal discussions:  

- We'll display a warning message inside Magento's admin to notify the user that the sandbox is enabled.
- We're allowing all functionality.  All API endpionts have been confirmed to either work or fail gracefully.

Model/Transaction/Refund.php
```diff
@@ -117,10 +117,6 @@ public function build(
     public function push($forceMethod = null) {
         $refundUpdatedAt = $this->originalRefund->getUpdatedAt();
         $refundSyncedAt = $this->originalRefund->getTjSalestaxSyncDate();
-        $refundApiKey = trim($this->scopeConfig->getValue(TaxjarConfig::TAXJAR_APIKEY,
-            \Magento\Store\Model\ScopeInterface::SCOPE_STORE,
-            $this->originalOrder->getStoreId()
```
fastdivision, line 122: @ryanreeves-taxjar Don't we still have to pass this store ID to get the correct API token for a refund's associated order?

Model/Transaction/Order.php
```diff
@@ -79,10 +79,6 @@ public function build(
     public function push($forceMethod = null) {
         $orderUpdatedAt = $this->originalOrder->getUpdatedAt();
         $orderSyncedAt = $this->originalOrder->getTjSalestaxSyncDate();
-        $orderApiKey = trim($this->scopeConfig->getValue(TaxjarConfig::TAXJAR_APIKEY,
-            \Magento\Store\Model\ScopeInterface::SCOPE_STORE,
-            $this->originalOrder->getStoreId()
```
fastdivision, line 84: @ryanreeves-taxjar Don't we still have to pass this store ID to get the correct API token for an order?

Observer/ConfigReview.php
```diff
@@ -57,28 +57,36 @@ class ConfigReview implements ObserverInterface
...
+    protected $taxjarConfig;
+
     /**
      * @param \Magento\Framework\App\Request\Http $request
      * @param CacheInterface $cache
      * @param ManagerInterface $eventManager
      * @param \Magento\Framework\Message\ManagerInterface $messageManager
      * @param ScopeConfigInterface $scopeConfig
      * @param NexusFactory $nexusFactory
+     * @param TaxjarConfig $taxjarConfig;
```
fastdivision, line 72: Semi-colon on the end here

Observer/ConfigReview.php
```diff
@@ -134,4 +143,17 @@ private function _reviewAddressValidation()
...
+
+    /**
+     * @return void
+     * @SuppressWarnings(Generic.Files.LineLength.TooLong)
+     */
+    private function _reviewSandboxMode()
+    {
+        if ($this->taxjarConfig->isSandboxEnabled()) {
+            // @codingStandardsIgnoreStart
+            $this->messageManager->addWarningMessage(__('Sandbox mode is enabled - this mode should only be used for testing.  Some features may not work as expected.'));
```
fastdivision, line 155: Think we should be clearer here around sandbox for calculations and staging environments:

```
Sandbox mode is enabled for sales tax calculations. We recommend only using this mode in a staging environment. Some features may not work as expected.
```

Can we add HTML here to link to our sandbox environment docs?

```html
Sandbox mode is enabled for sales tax calculations. We recommend only using this mode in a staging environment. Some features may not work as expected. <a href="https://developers.taxjar.com/api/reference/#sandbox-environment" target="_blank" rel="noopener">Learn more</a>
```

Model/Client.php
```diff
@@ -243,6 +251,9 @@ private function _getApiUrl($resource)
             case 'customers':
                 $apiUrl .= '/customers';
                 break;
+            case 'taxes':
+                $apiUrl .= '/magento/taxes';
+                break;
```
fastdivision, line 256: @ryanreeves-taxjar Do we still need this code if we're not moving forward with `Model/Client.php` for calculations?

Model/Logger.php
```diff
@@ -140,6 +147,10 @@ public function log($message, $label = '')
                     $label = '[' . strtoupper($label) . '] ';
                 }
 
+                if ($this->taxjarConfig->isSandboxEnabled()) {
+                    $label .= '[SANDBOX] ';
```
fastdivision, line 151: @ryanreeves-taxjar Can you provide sample output from the logs here?

Model/Smartcalcs.php
```diff
@@ -215,7 +220,8 @@ public function getTaxForOrder(
 
         if ($this->_orderChanged($order)) {
             $client = $this->clientFactory->create();
-            $client->setUri(TaxjarConfig::TAXJAR_API_URL . '/magento/taxes');
+
```
fastdivision, line 223: Extra newline here

fastdivision (changes requested):

etc/adminhtml/system.xml
fastdivision: @ryanreeves-taxjar Can we update the description here to include some more detail similar to the new notification?

```
Use the TaxJar sandbox environment to test API sales tax calculations. We recommend only using this mode in a staging environment. Some features may not work as expected.
```

etc/adminhtml/system.xml
```diff
@@ -73,7 +73,25 @@
                         <field id="debug">1</field>
                     </depends>
                 </field>
-                <field id="connected" translate="label" sortOrder="12" type="hidden" showInDefault="1" showInWebsite="1" showInStore="1">
+                <field id="sandbox" translate="label" sortOrder="12" type="select" showInDefault="1" showInWebsite="1" showInStore="1">
+                    <label>Enable Sandbox</label>
+                    <comment>Change all API requests to be made to TaxJar's sandbox account.</comment>
```
fastdivision, line 78: @ryanreeves-taxjar Can we update the description here to include some more detail similar to the new notification?

```
Use the TaxJar sandbox environment to test API sales tax calculations. We recommend only using this mode in a staging environment. Some features may not work as expected.
```

Model/Logger.php
ryanreeves-taxjar: sandbox_logs

Model/Transaction/Refund.php
ryanreeves-taxjar: You're 100% correct.  I got carried away refactoring and forgot that we need to pass the storeId here.

Model/Client.php
ryanreeves-taxjar: @fastdivision All of our api endpoints are defined here.  Even though `/magento/taxes` is (now) unused, I think it makes sense to keep it.

i18n/en_US.csv
```diff
@@ -111,3 +111,5 @@ Region,Region
 "API Token","API Token"
 "TaxJar Category Code","TaxJar Category Code"
 "Please redeploy production mode in the Magento CLI ($ bin/magento deploy:mode:set production) to ensure address validation works correctly.","Please redeploy production mode in the Magento CLI ($ bin/magento deploy:mode:set production) to ensure address validation works correctly."
+"Sandbox mode is enabled for sales tax calculations. We recommend only using this mode in a staging environment. Some features may not work as expected.","Sandbox mode is enabled for sales tax calculations. We recommend only using this mode in a staging environment. Some features may not work as expected."
+"Learn more","Learn more"
```
fastdivision, line 115: @ryanreeves-taxjar Should we update `en_US.csv` to cover config descriptions / comments as well?

Model/Logger.php
fastdivision: @ryanreeves-taxjar Can we have `[SANDBOX]` before `[POST]` and `[API]` so it lines up and makes it easier to scan?

Model/Transaction.php
```diff
@@ -96,6 +103,8 @@ public function __construct(
         $this->logger = $logger->setFilename(TaxjarConfig::TAXJAR_TRANSACTIONS_LOG);
         $this->objectManager = $objectManager;
         $this->helper = $helper;
+        $this->taxjarConfig = $taxjarConfig;
+        $this->apiKey = $this->taxjarConfig->getApiKey();
```
fastdivision, line 107: @ryanreeves-taxjar Do we use the API token anywhere in this class or through the inherited `Order` and `Refund` classes? Otherwise can we drop this code in `Model/Transaction.php`?

Model/Logger.php
ryanreeves-taxjar: 👍

fastdivision (approved):

PR #140, opened 2020-06-19
 Change to/from streets to only use the first street value

ryanreeves-taxjar:
### Context

The "from_street" address was created by concatenating the first and second street inputs.  This caused addresses with a second line to be improperly formatted.   

Screen Shot 2020-06-19 at 5 16 43 PM

### Description

This PR forces all to and from street addresses to only use the first input of the street, formatting the address like this:  
    **10 W 14th Ave Parkway**

### Performance

This PR has no impact on performance.  

### Testing

1. Configure the shipping origin address with a value in the second street input.
2. Sync a transaction to TaxJar
3. Confirm via API logs that the from_street value is correctly formatted. 

#### Versions

- [X] Magento 2.3
- [X] Magento 2.2
- [X] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -162,7 +162,7 @@ class Transaction
             'from_zip' => $fromPostcode,
             'from_state' => $fromState,
             'from_city' => $fromCity,
-            'from_street' => $fromStreet1 . $fromStreet2
+            'from_street' => $fromStreet1 . ', ' . $fromStreet2
         ];
     }
 
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -148,11 +148,7 @@ class Transaction
             \Magento\Store\Model\ScopeInterface::SCOPE_STORE,
             $order->getStoreId()
         );
-        $fromStreet1 = $this->scopeConfig->getValue('shipping/origin/street_line1',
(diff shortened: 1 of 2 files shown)


Model/Transaction.php
```diff
@@ -162,7 +162,7 @@ protected function buildFromAddress(
             'from_zip' => $fromPostcode,
             'from_state' => $fromState,
             'from_city' => $fromCity,
-            'from_street' => $fromStreet1 . $fromStreet2
+            'from_street' => $fromStreet1 . ', ' . $fromStreet2
```
dallendalton, line 165: Any need for a check to see if $fromStreet2 actually has a value? I don't think we want to end up with a comma on the end of $fromStreet1 when $fromStreet2 is empty.

ryanreeves-taxjar: Good catch @dallendalton!

dallendalton (dismissed):

codymullins (dismissed):

fastdivision (approved):

dallendalton (approved):

PR #149, opened 2020-08-05
Fixing `core_config_data` doesn't exist when installing

JosephMaxwell:
### Context
When installing Magento (after the TaxJar module has already been installed via composer, example is a build environment), we get this error:
```
Base table or view not found: 1146 Table ‘build_db.core_config_data’ doesn’t exist, query was: SELECT `main_table`.* FROM `core_config_data` AS `main_table`
```

### Description
This happens because the constructor fetched the information (ultimately from `core_config_data`) before this table was created. I've found that constructors are not an ideal place to fetch information, for this reason. The class is initialized before the schema creation happens.

### Performance
It now works :).

### Testing

1. Use composer to require Magento
2. Use composer to require the TaxJar module
3. Then run `bin/magento setup:install`.

#### Versions

- [ ] Magento 2.4
- [x] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [x] Magento Open Source (CE)
- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [x] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Observer/ImportCategories.php
+++ b/Observer/ImportCategories.php
@@ -99,7 +99,15 @@ class ImportCategories implements ObserverInterface
         $this->categoryFactory = $categoryFactory;
         $this->categoryResourceModel = $categoryResourceModel;
         $this->taxjarConfig = $taxjarConfig;
-        $this->apiKey = $this->taxjarConfig->getApiKey();
+    }
+    
+    private function getApiKey()
+    {
+        if (!$this->apiKey) {
+            $this->apiKey = $this->taxjarConfig->getApiKey();
+        }
+        
+        return $this->apiKey;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ryanreeves-taxjar:
Thanks @JosephMaxwell, and thanks for your contribution to our extension!  To confirm, you receive this error when requiring the TaxJar extension before installing Magento?

JosephMaxwell:
@ryanreeves-taxjar . The error occurs when I run these commands:
```
composer require magento/product-community-edition #something like this
composer require taxjar/module-taxjar
bin/magento setup:install
```

When Taxjar is installed _at the same time as Magento_ we get this error.

ryanreeves-taxjar:
Thanks @JosephMaxwell!  I was able to confirm this issue locally and we'll be adding this to our pre-release test cases.  Look for a review on your PR shortly!

ryanreeves-taxjar:
@JosephMaxwell I've reworked your PR somewhat in https://github.com/taxjar/taxjar-magento2-extension/pull/150.  Feel free to check it out and let us know what you think!

PR #152, opened 2020-08-19
Version 1.5.6

ryanreeves-taxjar:
- Fix support for table name prefixes used in admin grids
- Fix requiring TaxJar with composer before installing Magento (thanks @JosephMaxwell)
- Fix a fatal error that occasionally occurred when adding products to the cart (2.4 only)


JosephMaxwell:
Thank you @ryanreeves-taxjar.

PR #153, opened 2020-08-22
Add product tax code product attribute

ryanreeves-taxjar:
### Context

This PR adds a new product attribute called "tj_ptc" that is used to store a product tax code.  When set, this PTC is sent to the API.  When undefined, the tax class is used as a fallback.
Note:  Because PTCs can be assigned to any product:
- configurable products always use the PTC of the simple product added to the cart
- bundle products always use the PTCs of the selected simple products

### Description

Screen Shot 2020-09-01 at 11 42 42 AM

### Performance

This PR has a negligible impact on performance.

### Testing

1. Setup a default Magento installation and confirm all setup scripts have run (bin/magento setup:upgrade)
2. Add a PTC to a product of each type:
- single product
- configurable product (and child)
- bundle product
3. Confirm the PTCs are displayed on the product grid
Screen Shot 2020-08-21 at 5 51 49 PM

4. Place an order with each of the above products and confirm that the correct PTC is used.
5. Invoice and ship an order to confirm the correct PTC is used.
6. Issue a credit memo for the above order and confirm the correct PTC is used.  

Screen Shot 2020-08-21 at 6 08 49 PM

#### Versions

- [X] Magento 2.4
- [X] Magento 2.3
- [X] Magento 2.2
- [X] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Model/Tax/Sales/Total/Quote/Tax.php
+++ b/Model/Tax/Sales/Total/Quote/Tax.php
@@ -50,6 +50,16 @@ class Tax extends \Magento\Tax\Model\Sales\Total\Quote\Tax
     protected $taxCalculation;
 
     /**
+     * @var \Magento\Catalog\Api\ProductRepositoryInterface
+     */
+    protected $productRepository;
+
+    /**
+     * @var \Taxjar\SalesTax\Model\Logger
+     */
+    protected $logger;
+
+    /**
      * @param \Magento\Tax\Model\Config $taxConfig
      * @param \Magento\Tax\Api\TaxCalculationInterface $taxCalculationService
      * @param \Magento\Tax\Api\Data\QuoteDetailsInterfaceFactory $quoteDetailsDataObjectFactory
@@ -63,6 +73,8 @@ class Tax extends \Magento\Tax\Model\Sales\Total\Quote\Tax
      * @param \Magento\Framework\Pricing\PriceCurrencyInterface $priceCurrency,
      * @param \Magento\Tax\Api\Data\QuoteDetailsItemExtensionFactory $extensionFactory
      * @param \Taxjar\SalesTax\Model\Tax\TaxCalculation $taxCalculation
+     * @param \Magento\Catalog\Api\ProductRepositoryInterface $productRepository
+     * @param \Taxjar\SalesTax\Model\Logger $logger
      * @SuppressWarnings(PHPMD.ExcessiveParameterList)
      */
     public function __construct(
@@ -78,13 +90,17 @@ class Tax extends \Magento\Tax\Model\Sales\Total\Quote\Tax
         \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
         \Magento\Framework\Pricing\PriceCurrencyInterface $priceCurrency,
         \Magento\Tax\Api\Data\QuoteDetailsItemExtensionFactory $extensionFactory,
-        \Taxjar\SalesTax\Model\Tax\TaxCalculation $taxCalculation
+        \Taxjar\SalesTax\Model\Tax\TaxCalculation $taxCalculation,
+        \Magento\Catalog\Api\ProductRepositoryInterface $productRepository,
+        \Taxjar\SalesTax\Model\Logger $logger
     ) {
         $this->smartCalcs = $smartCalcs;
         $this->scopeConfig = $scopeConfig;
         $this->priceCurrency = $priceCurrency;
         $this->extensionFactory = $extensionFactory;
         $this->taxCalculation = $taxCalculation;
+        $this->productRepository = $productRepository;
+        $this->logger = $logger;
 
         parent::__construct(
             $taxConfig,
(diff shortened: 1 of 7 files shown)


Model/Tax/Sales/Total/Quote/Tax.php
```diff
@@ -328,6 +344,23 @@ public function mapItem(
...
+                $children = $item->getChildren();
+
+                if (is_array($children) && isset($children[0])) {
+                    $product = $this->productRepository->getById($children[0]->getProductId(), false, $item->getStore());
+                }
+            }
+
+            $extensionAttributes->setTjPtc($product->getTjPtc());
+        } catch (\Magento\Framework\Exception\NoSuchEntityException $e) {
+            $this->logger->log($e->getMessage());
```
fastdivision, line 361: What message would be returned here when the entity does not exist? Would it be helpful enough for the merchant to debug the issue and understand which product doesn't exist? Is it worth logging or better as a no-op?

Setup/UpgradeData.php
```diff
@@ -234,5 +234,38 @@ public function upgrade(
             $this->resourceConfig->deleteConfig('tax/taxjar/categories', 'default', '');
             $this->eventManager->dispatch('taxjar_salestax_import_categories');
         }
+
```
fastdivision, line 237: Extra newline here

Setup/UpgradeData.php
```diff
@@ -234,5 +234,38 @@ public function upgrade(
...
+            $code = 'tj_ptc';
+            $eavSetup = $this->eavSetupFactory->create(['setup' => $setup]);
+
+            $eavSetup->addAttribute(
+                \Magento\Catalog\Model\Product::ENTITY,
+                $code,
+                [
+                    'group' => 'General',
+                    'type' => 'text',
+                    'label' => 'TaxJar PTC',
```
fastdivision, line 249: I think consistency with the [tax class dropdown](https://www.taxjar.com/guides/integrations/magento2/#product-sales-tax-exemptions) works here: `TaxJar Category`. Can we display a basic dropdown for product categories similar to the Tax Class view so the merchant knows the PTC category name when reviewing the product?

Model/Tax/Sales/Total/Quote/Tax.php
ryanreeves-taxjar: The default error message is "Product does not exist".  We've seen similar errors in the past where products are deleted but still referenced.  

We can prefix it with a message of our own, but some error should appear in our logs to aid in debugging efforts.

fastdivision: @ryanreeves-taxjar I think it would be helpful if there was context around what specific line item has a product that does not exist. The deleted product ID may not be helpful since it's already deleted, but maybe the line ID / index and the line amount?

```
Product does not exist in line item `$itemDataObject->getCode()` for $`$item->getRowTotal()`
```

Model/Tax/Sales/Total/Quote/Tax.php
```diff
@@ -328,6 +344,23 @@ public function mapItem(
...
+
+            // Configurable products should use the PTC of the child (when available)
+            if ($product->getTypeId() == 'configurable') {
+                $children = $item->getChildren();
+
+                if (is_array($children) && isset($children[0])) {
+                    $product = $this->productRepository->getById($children[0]->getProductId(), false, $item->getStore());
+                }
+            }
+
```
dallendalton, line 358: Is there a scenario where we would want to use the parent product PTC instead of the child? For example if there is a TjPtc set on the parent but none set on the child? That way it would allow a single categorization to apply to all the child products. I'm thinking of scenarios where the child products are just variations on the parent, say for example T shirts, where the child products are just changes to size or color. In this case all would have the same PTC and the PTC wouldn't have to be applied individually.

Model/Transaction.php
```diff
@@ -276,7 +276,10 @@ protected function buildLineItems($order, $items, $type = 'order') {
 
             $product = $this->productRepository->getById($item->getProductId(), false, $order->getStoreId());
 
-            if ($product->getTaxClassId()) {
+            // Check for a PTC assigned directly to the product; otherwise fall back to tax classes
+            if ($product->getTjPtc()) {
+                $lineItem['product_tax_code'] = $product->getTjPtc();
+            } elseif ($product->getTaxClassId()) {
```
dallendalton, line 282: Just to confirm, the $product here would be the child in the case of a configurable product? So no need to pull the child products PTC here?

fastdivision:
Can we see an updated screenshot of the TaxJar Category dropdown? Thank you!

Model/Tax/Sales/Total/Quote/Tax.php
ryanreeves-taxjar: > Is there a scenario where we would want to use the parent product PTC instead of the child

No.  If a child PTC is defined, it supersedes the parent.  

> if there is a TjPtc set on the parent but none set on the child

The parent PTC is loaded via the Tax Class, but is only used when the child has no PTC.

Model/Transaction.php
ryanreeves-taxjar: @dallendalton Good catch.  I'd mistakenly assumed that we skip the configurable products and this would be a simple product, but the opposite is true.  Fix in the most recent commit.

codymullins:
@ryanreeves-taxjar let us know when this is ready for re-review.

PR #158, opened 2020-10-02
Fix mismatched tax class for configurable products

ryanreeves-taxjar:
### Context

When calculating taxes for configurable products, the tax class of the simple product was used during calculations but the tax class of the parent was used when reporting transactions.  

### Description

This PR standardizes tax classes to always use the tax class assigned to the simple product first, and if no tax class is set to fall back to the parent.

### Performance

n/a

### Testing

0.  Ensure logging is enabled.
1. Setup a configurable product with one tax class and a simple product with a different tax class.
2. Add the simple product to the cart and complete the checkout process.  Invoice and ship the order.
3. Review the logs to confirm the correct PTCs were used for calculations and transactions. 

#### Versions

- [X] Magento 2.4
- [X] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -274,14 +274,18 @@ class Transaction
                 'sales_tax' => $tax
             ];
 
+            // Load the tax class from the product.  For configurable products, check the child first
             $product = $this->productRepository->getById($item->getProductId(), false, $order->getStoreId());
+            $children = $item->getChildrenItems();
 
-            if ($product->getTaxClassId()) {
+            if (!empty($children) && $children[0]->getProduct()->getTaxClassId()) {
+                $taxClass = $this->taxClassRepository->get($children[0]->getProduct()->getTaxClassId());
+            } elseif ($product->getTaxClassId()) {
                 $taxClass = $this->taxClassRepository->get($product->getTaxClassId());
+            }
 
-                if ($taxClass->getTjSalestaxCode()) {
-                    $lineItem['product_tax_code'] = $taxClass->getTjSalestaxCode();
-                }
+            if ($taxClass && $taxClass->getTjSalestaxCode()) {
+                $lineItem['product_tax_code'] = $taxClass->getTjSalestaxCode();
             }
 
             $lineItems['line_items'][] = $lineItem;
--- a/Plugin/Tax/Model/Sales/Total/Quote/TaxjarCommonTaxCollector.php
+++ b/Plugin/Tax/Model/Sales/Total/Quote/TaxjarCommonTaxCollector.php
@@ -1,0 +1,59 @@
+<?php
+/**
+ * Taxjar_SalesTax
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Taxjar
+ * @package    Taxjar_SalesTax
+ * @copyright  Copyright (c) 2020 TaxJar. TaxJar is a trademark of TPS Unlimited, Inc. (http://www.taxjar.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+
+namespace Taxjar\SalesTax\Plugin\Tax\Model\Sales\Total\Quote;
+
+use Magento\ConfigurableProduct\Model\Product\Type\Configurable;
+use Magento\Quote\Model\Quote\Item\AbstractItem;
+use Magento\Tax\Api\Data\QuoteDetailsItemInterface;
+use Magento\Tax\Api\Data\QuoteDetailsItemInterfaceFactory;
+
+/**
+ * Plugin for CommonTaxCollector to apply Tax Class ID from child item for configurable product
+ */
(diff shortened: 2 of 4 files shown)


Plugin/Tax/Model/Sales/Total/Quote/TaxjarCommonTaxCollector.php
```diff
@@ -0,0 +1,59 @@
...
+        QuoteDetailsItemInterfaceFactory $itemDataObjectFactory,
+        AbstractItem $item
+    ) {
+        if ($item->getProduct()->getTypeId() === Configurable::TYPE_CODE) {
+            $taxClass = (int) $result->getTaxClassKey()->getValue();
+
+            // If a configurable product has no tax class, attempt to load it from the product
+            // This allows a simple product to inherit the tax class of it's configurable parent
+            if ($taxClass === 0) {
+                $result->getTaxClassKey()->setValue($item->getProduct()->getTaxClassId());
```
fastdivision, line 53: @ryanreeves-taxjar Is it possible to load the tax class from the item product here within `Model/SmartCalcs.php`? Do we need the tax collector plugin to do this successfully? Any concerns around this code executing if TaxJar is not enabled for calculations?

Model/Transaction.php
```diff
@@ -274,14 +274,18 @@ protected function buildLineItems($order, $items, $type = 'order') {
                 'sales_tax' => $tax
             ];
 
+            // Load the tax class from the product.  For configurable products, check the child first
             $product = $this->productRepository->getById($item->getProductId(), false, $order->getStoreId());
+            $children = $item->getChildrenItems();
 
-            if ($product->getTaxClassId()) {
+            if (!empty($children) && $children[0]->getProduct()->getTaxClassId()) {
+                $taxClass = $this->taxClassRepository->get($children[0]->getProduct()->getTaxClassId());
```
fastdivision, line 282: @ryanreeves-taxjar Does this have any impact on how we'd handle child items for other types of products such as bundle products or just configurable? When calling `getChildrenItems`, do configurable products always return one simple child product? Should we add a check for configurable product type?

Plugin/Tax/Model/Sales/Total/Quote/TaxjarCommonTaxCollector.php
ryanreeves-taxjar: **> Is it possible to load the tax class from the item product within Model/SmartCalcs.php instead?  **
The Quote\ItemDetails instance (that we iterate over in Smartcalcs _getLineItems) has no knowledge of the actual product.  My initial approach was to add the product id so we could look it up in SmartCalcs, but after reviewing the Configurable Product CommonTaxCollector, I felt like this was a cleaner approach.  

https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/ConfigurableProduct/Plugin/Tax/Model/Sales/Total/Quote/CommonTaxCollector.php


**>  Do we need the tax collector plugin to do this successfully?**
We need to extend Quote\ItemDetails in some way.  The plugin seems to be the cleanest but I'm open to alternatives.  

**> Any concerns around this code executing if TaxJar is not enabled for calculations?**

If we disable this plugin when rates calculations are disabled, we'll see a mismatch between the tax class used during checkout and during transaction sync.  This may not actually be a problem, since rates aren't being calculated by TJ, but I think it's better to be consistent.

Model/Transaction.php
ryanreeves-taxjar: Because of the exclusion checks that take place before this code (skipping certain items), there are no conflicts here.  I don't see any problems adding a check for configurable products, but it's not necessary.

Plugin/Tax/Model/Sales/Total/Quote/TaxjarCommonTaxCollector.php
fastdivision: @ryanreeves-taxjar Thanks for the info. Have we considered extension attributes to get back the product tax class ID within [/Model/Tax/Sales/Total/Quote/Tax.php#mapItem](https://github.com/taxjar/taxjar-magento2-extension/blob/develop/Model/Tax/Sales/Total/Quote/Tax.php#L292-L293) and [defined here](https://github.com/taxjar/taxjar-magento2-extension/blob/develop/etc/extension_attributes.xml#L20)?

ryanreeves-taxjar: If we modify the tax class inside the `mapItem` function, it will be overwritten by the `afterMapItem` function in ConfigurableProduct.

Your suggested approach (set an extension attribute and add a check inside SmartCalcs) would also work.  To me it's cleaner to modify the tax class directly, but I'm open to alternative approaches if I'm overlooking something.

fastdivision: Good to know, thanks @ryanreeves-taxjar. I'm open to either approach. This approach seems the cleanest. The other approach may be easier to review / maintain in the future if it lives in SmartCalcs. Perhaps additional detail in the plugin class docblock above would be helpful to explain where this code is being used in SmartCalcs.php and the purpose behind it specifically for our integration.

fastdivision (dismissed):
Looks good to me 👍

PR #162, opened 2020-10-27
Persist PTCs on OrderItems

ryanreeves-taxjar:
### Context

This PR adds an extra column (tj_ptc) to the sales_order_item table.  It saves the PTC for each OrderItem during creation via the ItemRepository beforeSave plugin.

### Description

Product tax codes can change between an order being placed and the order being synced to TaxJar.  By saving the PTC on the sales_order_item table, we can ensure the correct code is synced.  

Screen Shot 2020-10-26 at 7 42 28 PM

### Performance

Before methods in plugins have a negligible impact on performance.

### Testing

1. Obtain this PR and install it by running `bin/magento setup:upgrade`
2. Confirm that the "tj_ptc" column is added to the table `sales_order_item`
3. Place an order with product(s) with PTCs assigned
4. Confirm that the new order items contain the appropriate PTCs
5. Change the PTC on one of the products ordered
6. Invoice and Ship the order
7. Confirm that the correct PTC(s) were synced to TaxJar

#### Versions

- [X] Magento 2.4
- [X] Magento 2.3
- [] Magento 2.2
- [] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -375,7 +375,8 @@ class Transaction
     {
         // Check for a PTC saved to the Item
         if ($item->getTjPtc()) {
-            return $item->getTjPtc();
+            // Return the PTC, or an empty string if the TAXABLE_TAX_CODE is present
+            return $item->getTjPtc() != $this->taxjarConfig::TAXJAR_TAXABLE_TAX_CODE ?: '';
         }
 
         // If no PTC is saved on the Item, attempt to load it from the product or tax class
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -374,9 +374,15 @@ class Transaction
     protected function getProductTaxCode($item, $order)
     {
         // Check for a PTC saved to the Item
+        // For configurable products, load the PTC of the child
+        if ($item->getHasChildren() && $item->getProductType() == 'configurable') {
+            $child = $item->getChildrenItems()[0];
+            return $child->getTjPtc() != $this->taxjarConfig::TAXJAR_TAXABLE_TAX_CODE ? $child->getTjPtc() : '';
+        }
+
         if ($item->getTjPtc()) {
             // Return the PTC, or an empty string if the TAXABLE_TAX_CODE is present
-            return $item->getTjPtc() != $this->taxjarConfig::TAXJAR_TAXABLE_TAX_CODE ?: '';
+            return $item->getTjPtc() != $this->taxjarConfig::TAXJAR_TAXABLE_TAX_CODE ? $item->getTjPtc() : '';
         }
 
         // If no PTC is saved on the Item, attempt to load it from the product or tax class
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -373,6 +373,12 @@ class Transaction
      */
     protected function getProductTaxCode($item, $order)
     {
+        // Check for a PTC saved to the Item
+        if ($item->getTjPtc()) {
+            return $item->getTjPtc();
+        }
+
+        // If no PTC is saved on the Item, attempt to load it from the product or tax class
         try {
             $product = $this->productRepository->getById($item->getProductId(), false, $order->getStoreId());
 
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -376,8 +376,12 @@ class Transaction
         // Check for a PTC saved to the Item
         // For configurable products, load the PTC of the child
         if ($item->getHasChildren() && $item->getProductType() == 'configurable') {
-            $child = $item->getChildrenItems()[0];
-            return $child->getTjPtc() != $this->taxjarConfig::TAXJAR_TAXABLE_TAX_CODE ? $child->getTjPtc() : '';
+            $children = $item->getChildrenItems();
+
+            if (!empty($children) && is_array($children)) {
+                $child = reset($children);
+                return $child->getTjPtc() != $this->taxjarConfig::TAXJAR_TAXABLE_TAX_CODE ? $child->getTjPtc() : '';
+            }
         }
 
         if ($item->getTjPtc()) {
--- a/Plugin/Sales/Model/Order/ItemRepository.php
+++ b/Plugin/Sales/Model/Order/ItemRepository.php
@@ -1,0 +1,53 @@
+<?php
+/**
+ * Taxjar_SalesTax
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Taxjar
+ * @package    Taxjar_SalesTax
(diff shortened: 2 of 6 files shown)


Plugin/Sales/Model/Order/ItemRepository.php
```diff
@@ -0,0 +1,53 @@
...
+     * @param \Magento\Sales\Model\Order\ItemRepository $subject
+     * @param \Magento\Sales\Api\Data\OrderItemInterface $item
+     */
+    public function beforeSave(
+        \Magento\Sales\Model\Order\ItemRepository $subject,
+        \Magento\Sales\Api\Data\OrderItemInterface $item
+    ) {
+        // Only set the product tax code when first creating the OrderItem
+        if ($item->getItemId() === null) {
+            $product = $this->productRepository->getById($item->getProductId());
```
fastdivision, line 47: @ryanreeves-taxjar Any potential edge cases where the order item would not have a product before creation? Should we check if `$product` exists to be safe?

ryanreeves-taxjar: Deleted products are automatically removed from the quote.  It shouldn't be possible to reach this step with a product that was deleted.  

I'll go ahead and add a catch for NoSuchEntityExceptions anyways.

Model/Transaction.php
```diff
@@ -373,6 +373,12 @@ protected function getParentAmounts($attr, $items, $type = 'order') {
      */
     protected function getProductTaxCode($item, $order)
     {
+        // Check for a PTC saved to the Item
+        if ($item->getTjPtc()) {
```
fastdivision, line 377: @ryanreeves-taxjar If a product is set to the "Fully Taxable" category and we save an empty string in the `tj_ptc` column, will this conditional be truthy and return back an empty tax code? Or will it fall back to looking up the tax code on the product? Checking to ensure an item that goes from "Fully Taxable" to another category before transaction sync remains fully taxable.

ryanreeves-taxjar: @jakedjohnson NULL values are stored when no PTC is present, which does cause `if ($item->getTjPtc())` to equate to false.  This causes a potentially incorrect second lookup (the behavior I mentioned in chat).  

One solution is to save a non-empty value that represents no PTC.  I don't think we have a "Fully Taxable" PTC, so maybe 00000 as the opposite of 99999?  Then:

- If the column is empty, the order_item was created before the `tj_ptc` field existed and fall back to the product
- If it's not empty, only use the 'tj_ptc' field

Thoughts?

fastdivision: @ryanreeves-taxjar Feel free to use `11111` for the "Fully Taxable" category if that helps prevent any issues here. It's equivalent to not passing a tax code or a blank string.

Model/Transaction.php
```diff
@@ -373,6 +373,12 @@ protected function getParentAmounts($attr, $items, $type = 'order') {
      */
     protected function getProductTaxCode($item, $order)
     {
+        // Check for a PTC saved to the Item
```
codymullins, line 376: Does `getTjPtc()` hit the database? It might be cleaner to just assign the result of the method to a variable so we don't need to call it twice.

Plugin/Sales/Model/Order/ItemRepository.php
```diff
@@ -0,0 +1,71 @@
...
+    public function beforeSave(
+        \Magento\Sales\Model\Order\ItemRepository $subject,
+        \Magento\Sales\Api\Data\OrderItemInterface $item
+    ) {
+        // Only set the product tax code when first creating the OrderItem
+        if ($item->getItemId() === null) {
+            try {
+                $product = $this->productRepository->getById($item->getProductId());
+                $item->setTjPtc($product->getTjPtc());
+            } catch (NoSuchEntityException $e) {
```
codymullins, line 63: What does this mean: `possibly missing PTCs on OrderItems.`?

Setup/UpgradeSchema.php
```diff
@@ -150,5 +150,27 @@ public function upgrade(SchemaSetupInterface $setup, ModuleContextInterface $con
...
+
+            /**
+             * Update table 'sales_order_item'
+             */
+            $installer->getConnection()->addColumn(
+                $installer->getTable('sales_order_item'),
+                'tj_ptc',
+                [
+                    'type' => \Magento\Framework\DB\Ddl\Table::TYPE_TEXT,
+                    'length' => 255,
```
codymullins, line 166: We don't have any PTCs as long as 255 characters, do we? I'm not familiar with the database behind this, but do we want to size this column to fit? We could still give ourselves a little wiggle room if we would like, but 255 seems really high.

ryanreeves-taxjar: You're right; we use 32 elsewhere.  I'll update it!

Plugin/Sales/Model/Order/ItemRepository.php
ryanreeves-taxjar: If a product had a PTC assigned, but the product was deleted, that PTC won't be saved.  OR there was never a PTC to begin with.  With the product deleted, we don't know.

Model/Transaction.php
ryanreeves-taxjar: It does not; it uses a local cache on the object.  https://gielberkers.com/magentos-magical-getters-setters-demystified/

Setup/UpgradeSchema.php
ryanreeves-taxjar: Screen Shot 2020-10-28 at 6 05 49 PM

Model/Transaction.php
```diff
@@ -373,6 +373,13 @@ protected function getParentAmounts($attr, $items, $type = 'order') {
...
+        if ($item->getTjPtc()) {
+            // Return the PTC, or an empty string if the TAXABLE_TAX_CODE is present
+            return $item->getTjPtc() != $this->taxjarConfig::TAXJAR_TAXABLE_TAX_CODE ?: '';
```
fastdivision, line 379: @ryanreeves-taxjar Does this line return a boolean value instead of the PTC or taxable tax code?

ryanreeves-taxjar: No, it's a shorthand ternary operator.

ryanreeves-taxjar:
Screen Shot 2020-10-30 at 5 37 41 PM
[order-77.log](https://github.com/taxjar/taxjar-magento2-extension/files/5468150/order-77.log)

Model/Transaction.php
```diff
@@ -373,6 +373,19 @@ protected function getParentAmounts($attr, $items, $type = 'order') {
      */
     protected function getProductTaxCode($item, $order)
     {
+        // Check for a PTC saved to the Item
+        // For configurable products, load the PTC of the child
+        if ($item->getHasChildren() && $item->getProductType() == 'configurable') {
+            $child = $item->getChildrenItems()[0];
```
fastdivision, line 379: @ryanreeves-taxjar Do we need to use `reset` here similar to #165?

ryanreeves-taxjar: I think that's a good idea.

fastdivision (approved):

PR #165, opened 2020-11-09
Fix array lookup for configurable product PTCs

ryanreeves-taxjar:
### Context

Some users are still experiencing fatal errors when trying to load the PTC from the children of configurable products.

### Description

Some extensions change the array keys of the children of configurable products.  This PR uses `array_first` to get the first element in the array.

### Performance

This PR has a negligible impact on performance.  

### Testing

1. Assign a PTC to a parent configurable product, and a different PTC to the child.  
2. Checkout with the product in the cart, and invoice/ship the order.
3. Confirm that the correct PTC synced to TaxJar

#### Versions

- [X] Magento 2.4
- [X] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -386,8 +386,12 @@ class Transaction
                 $children = $item->getChildrenItems();
             }
 
-            if (!empty($children) && $children[0]->getProduct()->getTaxClassId()) {
-                $taxClass = $this->taxClassRepository->get($children[0]->getProduct()->getTaxClassId());
+            if (!empty($children) && is_array($children)) {
+                $child = array_first($children);
+
+                if ($child->getProduct()->getTaxClassId()) {
+                    $taxClass = $this->taxClassRepository->get($child->getProduct()->getTaxClassId());
+                }
             } elseif ($product->getTaxClassId()) {
                 $taxClass = $this->taxClassRepository->get($product->getTaxClassId());
             }
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -387,7 +387,7 @@ class Transaction
             }
 
             if (!empty($children) && is_array($children)) {
-                $child = array_first($children);
+                $child = reset($children);
 
                 if ($child->getProduct()->getTaxClassId()) {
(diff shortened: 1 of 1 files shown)


Model/Transaction.php
```diff
@@ -386,8 +386,12 @@ protected function getProductTaxCode($item, $order)
                 $children = $item->getChildrenItems();
             }
 
-            if (!empty($children) && $children[0]->getProduct()->getTaxClassId()) {
-                $taxClass = $this->taxClassRepository->get($children[0]->getProduct()->getTaxClassId());
+            if (!empty($children) && is_array($children)) {
+                $child = array_first($children);
```
dbernheisel, line 390: Just so I understand (I don't write PHP often) -- where does `array_first` come from? My google-fu seems to only show this as a Laravel thing.

ryanreeves-taxjar: @dbernheisel This is part of the `weew/helpers-array` package that's included with Magento.  
https://packagist.org/packages/weew/helpers-array

If you're worried about additional dependencies, we can rework it to use php's `reset` instead.  I chose `array_first` because it's a bit more intuitive.  
https://www.php.net/manual/en/function.reset.php

fastdivision: @ryanreeves-taxjar Is the `weew/helpers-array` package still included in Magento 2.3+? Seeing it in 2.2 but not 2.3 or 2.4: https://devdocs.magento.com/guides/v2.4/release-notes/packages-open-source.html

I agree with using a native PHP method such as `reset` to avoid regressions in the future if Magento decides to remove a package from core.

ryanreeves-taxjar: It looks like it was moved to the `magento2-functional-testing-framework` package, which is an optional dev dependency.  Updated to use `reset`.  

https://github.com/magento/magento2-functional-testing-framework/blob/develop/composer.json

fastdivision (approved):

dbernheisel (approved):

PR #166, opened 2020-11-11
Handling adjustment fees for $0 refunds

ryanreeves-taxjar:
### Context

$0 orders with a shipping amount encounter a "division by zero" error when attempting to sync them to TaxJar.

### Description

When distributing refund adjustment fees over the line items, we divide by the subtotal.  In the case where the subtotal is exactly 0, this causes an error.  This PR skips distributing adjustment fees over line items when the subtotal is 0.

Note: If an adjustment fee is defined, and the subtotal is $0, the adjustment fee will not be included.  However manually including the adjustment fee as a line item results in an incorrect expected tax amount.  I believe it's safe to ignore the adjustment fees in this situation.  

Screen Shot 2020-11-11 at 12 43 15 PM

### Performance

n/a

### Testing

1. Place an order with a $0 (free) product and a shipping amount.  Ship to a location with shipping tax (ex: NY)
2. Invoice and Ship the order.  Issue a credit memo for the same order, optionally including adjustment fees.
3. Confirm that the credit memo syncs to TaxJar correctly.

#### Versions

- [X] Magento 2.4
- [X] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [X] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [ ] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Model/Transaction/Refund.php
+++ b/Model/Transaction/Refund.php
@@ -80,13 +80,14 @@ class Refund extends \Taxjar\SalesTax\Model\Transaction
             $adjustmentFee = $creditmemo->getAdjustmentNegative();
             $adjustmentRefund = $creditmemo->getAdjustmentPositive();
 
-            if ($subtotal != 0) {
-                // Discounts on credit memos act as fees and shouldn't be included in $itemDiscounts
-                foreach ($this->request['line_items'] as $k => $lineItem) {
+            // Discounts on credit memos act as fees and shouldn't be included in $itemDiscounts
+            foreach ($this->request['line_items'] as $k => $lineItem) {
+                if ($subtotal != 0) {
                     $lineItemSubtotal = $lineItem['unit_price'] * $lineItem['quantity'];
                     $this->request['line_items'][$k]['discount'] += ($adjustmentFee * ($lineItemSubtotal / $subtotal));
-                    $itemDiscounts += $lineItem['discount'];
                 }
+
+                $itemDiscounts += $lineItem['discount'];
             }
 
             if ($adjustmentRefund) {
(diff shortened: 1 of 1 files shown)


Model/Transaction/Refund.php
```diff
@@ -80,11 +80,13 @@ public function build(
...
-            foreach ($this->request['line_items'] as $k => $lineItem) {
-                $lineItemSubtotal = $lineItem['unit_price'] * $lineItem['quantity'];
-                $this->request['line_items'][$k]['discount'] += ($adjustmentFee * ($lineItemSubtotal / $subtotal));
-                $itemDiscounts += $lineItem['discount'];
+            if ($subtotal != 0) {
+                // Discounts on credit memos act as fees and shouldn't be included in $itemDiscounts
+                foreach ($this->request['line_items'] as $k => $lineItem) {
+                    $lineItemSubtotal = $lineItem['unit_price'] * $lineItem['quantity'];
+                    $this->request['line_items'][$k]['discount'] += ($adjustmentFee * ($lineItemSubtotal / $subtotal));
+                    $itemDiscounts += $lineItem['discount'];
```
fastdivision, line 88: @ryanreeves-taxjar If a credit memo has a $0 subtotal, do we also want to skip this line on `$itemDiscounts`? This line was initially called prior to modifying discounts based on the adjustment fee. Should we only wrap line 86-87 in the conditional, or keep as-is? Only use case I can imagine this may impact would be a $0 subtotal credit memo with a shipping refund and a shipping discount.

fastdivision (approved):

PR #172, opened 2021-01-29
API Version Header

dallendalton:
### Context
Adding the x-api-version header will allow the extension to use the latest version of the TaxJar API.

### Description
This PR adds the x-api-version header to all TaxJar API Requests.

### Performance
N/A

### Testing
- Create a test order on the front end and confirm tax is calculated
- Put the order in complete/closed status and confirm order is synced to TaxJar

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3

- [X] Magento Commerce (EE)

- [X] PHP 7.x

### DIFF
--- a/Model/Client.php
+++ b/Model/Client.php
@@ -184,7 +184,10 @@ class Client
             'useragent' => $this->tjHelper->getUserAgent(),
             'referer' => $this->tjHelper->getStoreUrl()
         ]);
-        $client->setHeaders('Authorization', 'Bearer ' . $this->apiKey);
+        $client->setHeaders([
+            'Authorization' => 'Bearer ' . $this->apiKey,
+            'x-api-version' => TaxJarConfig::TAXJAR_X_API_VERSION
+        ]);
 
         return $client;
     }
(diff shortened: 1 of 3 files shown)


Model/Client.php
```diff
@@ -184,7 +184,10 @@ private function getClient($url, $method = \Zend_Http_Client::GET)
             'useragent' => $this->tjHelper->getUserAgent(),
             'referer' => $this->tjHelper->getStoreUrl()
         ]);
-        $client->setHeaders('Authorization', 'Bearer ' . $this->apiKey);
+        $client->setHeaders(array(
```
fastdivision: @dallendalton We can use shortened array syntax here `[]` given Magento's PHP requirements.

fastdivision (changes requested):

Model/Client.php
dallendalton: This is now fixed.

fastdivision (approved):

poops (approved):

PR #182, opened 2021-04-26
Address Validation Javascript Fix

dallendalton:
### Context
The address validation javascript is currently broken on Magento 2.4.

### Description
This PR fixes the issue by the checkout provider registry instead of the quote.

### Performance
N/A

### Testing
1. Ensure Address Validation is enabled in the TaxJar settings.
2. Add a product to the cart and proceed to checkout.
3. Enter an address, and then choose the selected address.
4. Proceed to the next step, and you will see that the selected suggested address was not applied to the cart.

After applying this PR, the selected address will apply to the cart.

#### Versions
- [X] Magento 2.4.2
- [X] Magento 2.3.6

### DIFF
--- a/view/base/web/js/model/address_validation_core.js
+++ b/view/base/web/js/model/address_validation_core.js
@@ -128,8 +128,6 @@ function (ko, $) {
         },
 
         addressMatches: function (address, addressToCompare) {
-            console.log(JSON.stringify(address));
-            console.log(JSON.stringify(addressToCompare));
             return JSON.stringify(address) === JSON.stringify(addressToCompare);
         }
     };
--- a/view/base/web/js/model/address_validation_core.js
+++ b/view/base/web/js/model/address_validation_core.js
@@ -106,25 +106,23 @@ function (ko, $) {
         },
 
         formatAddress: function (address) {
-            var formattedAddress = {
+            return {
                 'street0': address.street[0],
                 'city': address.city,
                 'region': address.region_id,
                 'country': address.country_id,
                 'postcode': address.postcode
             };
-            return formattedAddress;
         },
 
         formatSuggestedAddress: function (suggestedAddress) {
-            var formattedAddress = {
+            return {
                 'street0': suggestedAddress.street[0],
                 'city': suggestedAddress.city,
                 'region': suggestedAddress.regionId,
                 'country': suggestedAddress.countryId,
                 'postcode': suggestedAddress.postcode
             };
-            return formattedAddress;
         },
 
         addressMatches: function (address, addressToCompare) {
--- a/view/base/web/js/model/address_validation_core.js
+++ b/view/base/web/js/model/address_validation_core.js
@@ -24,6 +24,8 @@ function (ko, $) {
     return {
         suggestedAddresses: ko.observable([]),
         activeAddress: {},
+        validatedAddress: {},
+        isRefresh: false,
 
         getAddressValidationUrl: function () {
             return '/rest/V1/Taxjar/address_validation/';
@@ -33,7 +35,7 @@ function (ko, $) {
             var self = this;
 
             // Skip if non-US shipping address
-            if (addr && addr.countryId !== 'US') {
+            if (addr && addr.country_id !== 'US') {
                 self.updateSuggestedAddresses([]);
 
                 if (typeof onFail === 'function') {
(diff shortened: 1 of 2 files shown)


view/base/web/js/model/address_validation_core.js
```diff
@@ -93,8 +89,46 @@ function (ko, $) {
...
 
+        isSuggestedAddress: function (address) {
+            var self = this;
+            var isSuggested = false;
+            var suggestedAddresses = this.suggestedAddresses();
+            suggestedAddresses.forEach(function(suggestedAddress, index) {
+                if (self.addressMatches(self.formatAddress(address), self.formatSuggestedAddress(suggestedAddress.address))) {
+                    isSuggested = true;
+                }
+            });
```
poops, line 100: I'm not sure how large `suggestedAddresses` could be, so this may not matter, but what about returning early if a match is found? I think you'd have to use a `for` loop instead of `forEach` though.

view/base/web/js/model/address_validation_core.js
```diff
@@ -93,8 +89,46 @@ function (ko, $) {
...
+
+        formatAddress: function (address) {
+            var formattedAddress = {
+                'street0': address.street[0],
+                'city': address.city,
+                'region': address.region_id,
+                'country': address.country_id,
+                'postcode': address.postcode
+            };
+            return formattedAddress;
```
poops: With this and `formatSuggestedAddress`, could you return the object instead of instantiating a var?

poops (dismissed):
Left some minor feedback, but code LGTM!

view/base/web/js/model/address_validation_core.js
```diff
@@ -93,8 +89,46 @@ function (ko, $) {
...
+                'city': suggestedAddress.city,
+                'region': suggestedAddress.regionId,
+                'country': suggestedAddress.countryId,
+                'postcode': suggestedAddress.postcode
+            };
+            return formattedAddress;
+        },
+
+        addressMatches: function (address, addressToCompare) {
+            return JSON.stringify(address) === JSON.stringify(addressToCompare);
```
dbernheisel, line 129: objects are not ordered, so I don't think this is as reliable as we hope. I don't believe this will pass if the keys are somehow ordered differently since this compares strings.

```javascript
JSON.stringify({'one': 'foo', 'two': 'bar'}) === JSON.stringify({'two': 'bar', 'one': 'foo'})
// false
```

I'm not seeing a jQuery utility that helps with this. If we know the object is flat (looks like it is), then this could turn into an array of two-item arrays, sort by first item, and then check equality of those.

Some psuedocode to get started:

```javascript
sortAddress: function(addr) {
    var sorted = [];

    for(var i in addr) {
        sorted.push([i, addr[i]]);
    }

    return sorted.sort();
}

sortAddress(addr) == sortAddress(addressToCompare)
```

view/frontend/web/js/view/suggested_address_checkout_step.js
```diff
@@ -106,15 +114,34 @@ function (
...
+                    };
+                } else {
+                    originalAddress.taxjar_attributes = {
+                        suggestedAddress: false
+                    };
+                }
+
+                // street address does not update when updating the checkout provider shipping address
+                // that input must be updated individually
+                $('.form-shipping-address input[name="street[0]"]').val(originalAddress.street[0]);
```
dbernheisel, line 141: Feels dirty to directly reach into the form and update it. Is this a typical thing to do?  Is there a way to do this via Magento? (i'm not familiar with this generally works).

view/base/web/js/model/address_validation_core.js
dallendalton: I like the suggestion, and you are right about having to use a for loop.  However this array should only have have 2 or 3 entries. This whole feature is going to be refactored in the future, so I think for now the current approach is satisfactory.

view/frontend/web/js/view/suggested_address_checkout_step.js
dallendalton: I don't particularly like this approach either. However when updating through the "normal" magento way (the checkoutProvider.set line) it does not update the street address. None of the other methods I tested would correctly update the street address to I had to revert to the approach in this PR.

view/base/web/js/model/address_validation_core.js
dallendalton: Agreed, will get this updated.

view/base/web/js/model/address_validation_core.js
dallendalton: In this particular use case, because they same functions are used to format the address when they are stored, they will always have the same order. I think building a function to do a deep comparison isn't necessary here.

dbernheisel (approved):

poops (approved):

PR #186, opened 2021-05-21
Fix suggestion address issue

rangerz:
### Context
Fix some address validation issue.

### Description
- `shipping.js` - Use correct `country_id` and `region_id` keys because of the [commit changes](https://github.com/taxjar/taxjar-magento2-extension/commit/8aa43f025d4c3c26336f80696508aedb02b60d45#)
image001

- ~~`suggested_address_checkout_step.js` - add `waitForElement` function for `rearrangeSteps` function, because `.onestep-shipping-address` of `Aheadwork_OneStetCheckout` may render later when refresh page (ex: enter F5 key)~~ (CAN NOT Duplicate in out of box environment - reverted the changes)

- `address_validation_core.js` - add `isValidAddress` function with `address.postcode` condition, and also refactor code for nested if statements.

### Performance
None

### Testing
1. Add some product to cart and into the checkout page
2. Fill shipping address when complete with Street 1, City, Country (US), State, and Zip/Postal Code (It should enter a little incorrect for trigger address suggestion)
3. Show suggested address correctly
4. F5 (Refresh page), suggested address show correctly if an address is filled. - Aheadwork_OneStepCheckout only

#### Versions

- [x] Magento 2.4
- [ ] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [x] Magento Commerce (EE)
- [x] Magento B2B
- [x] Magento Cloud

- [x] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/view/base/web/js/model/address_validation_core.js
+++ b/view/base/web/js/model/address_validation_core.js
@@ -34,59 +34,59 @@ function (ko, $) {
         getSuggestedAddresses: function (addr, onDone, onFail) {
             var self = this;
 
-            // Skip if non-US shipping address
-            if (addr && addr.country_id !== 'US') {
-                self.updateSuggestedAddresses([]);
+            if (!this.isValidAddress(addr)) {
+                // Skip if non-US shipping address
+                if (addr && addr.country_id !== 'US') {
+                    self.updateSuggestedAddresses([]);
 
-                if (typeof onFail === 'function') {
-                    onFail('NON_US_SHIPPING_ADDRESS');
+                    if (typeof onFail === 'function') {
+                        onFail('NON_US_SHIPPING_ADDRESS');
+                    }
+                } else {
+                    if (typeof onFail === 'function') {
+                        onFail('MISSING_ADDRESS_FIELDS');
+                    }
                 }
                 return;
             }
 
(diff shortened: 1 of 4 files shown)


rangerz:
@dallendalton @poops @dbernheisel 
Here needs some reviewers for fixing issue in #182

view/base/web/js/model/address_validation_core.js
```diff
@@ -105,6 +105,17 @@ function (ko, $) {
             this.suggestedAddresses(addr);
         },
 
+        isValidAddress: function (address) {
+            return !!(
+                address &&
+                address.country_id === 'US' &&
```
dallendalton: This method needs to only validate that the country_id is set, not that it is "US". In it's current state it prevents the logic that validates the country_id is US from ever being reached.

dallendalton (changes requested):
I've been testing this and it looks good. Just the one minor change is needed in the isValidAddress method and then we will get it merged in. Thanks!

view/base/web/js/model/address_validation_core.js
rangerz: You are correct. I updated.

poops (approved):

dallendalton (approved):

PR #191, opened 2021-06-30
Fix various address validation issues

dallendalton:
### Context
Address validation should be able to be used at multiple locations in Magento. These locations are: during guest checkout, during checkout by a logged in user (for both new addresses and saved addresses), in the address book (in account) for a logged in customer) and during order creation in the admin interface.

### Description
There are multiple issues with address validation:

1. While creating an order in admin, it is incorrectly displaying that it can't validate a non US address, even when the country is set to US.
2. When a logged in user checks out, saved addresses are not being validated.
3. When a logged in user checks out, after adding a new address (which is validated) and then switching to a saved address, the suggested addresses are not updated when the saved address is selected.
4. Buttons in the modals that are displayed when validating addresses don't give a clear idea of what they do.
5. Address validation is not working in the address book for logged in customers.

### Performance
No effect.

### Testing

1. Create an order in admin using a US address. Address validation should work correctly.
2. Add an item to the cart with a logged in customer. Address validation should work for both saved and new addresses.
3. Change addresses using the address book for a logged in customer, address validation should work properly.

#### Versions

- [X] Magento 2.4
- [X] Magento 2.3

### DIFF
--- a/view/adminhtml/web/js/suggested_addresses_admin.js
+++ b/view/adminhtml/web/js/suggested_addresses_admin.js
@@ -43,20 +43,46 @@ define([
 
         initialize: function (config) {
             this._super();
-
             var self = this;
+            this.addressModal = this.buildModal();
 
-            this.addressModal = $('#tj-suggested-addresses').modal({
+            if (config.controller === 'order_create') {
+                this.appendButtonToOrder();
+
+                if ('MutationObserver' in window) {
+                    var orderObserver = new MutationObserver(function (mutations) {
+                        var addressObserver = new MutationObserver(function (mutations) {
+                            self.appendButtonToOrder();
+                            orderObserver.disconnect();
+                        });
+
+                        self.appendButtonToOrder();
+
+                        if ($('.order-billing-address, .order-shipping-address').length) {
+                            addressObserver.observe($('.order-billing-address').get(0), {childList: true});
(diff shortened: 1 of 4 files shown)


view/adminhtml/web/js/suggested_addresses_admin.js
```diff
@@ -43,20 +43,46 @@ define([
...
+                            self.appendButtonToOrder();
+                            orderObserver.disconnect();
+                        });
+
+                        self.appendButtonToOrder();
+
+                        if ($('.order-billing-address, .order-shipping-address').length) {
+                            addressObserver.observe($('.order-billing-address').get(0), {childList: true});
+                            addressObserver.observe($('.order-shipping-address').get(0), {childList: true});
+                        }
```
poops: Could this be split out into 2 separate `if` statements? Or is that `observe` method safe to call on a non-existent element?

poops (dismissed):
One minor "observation", but otherwise LGTM!

view/adminhtml/web/js/suggested_addresses_admin.js
dallendalton: I tested this out and you were correct that it should have been split into two if statements. It has now been fixed.

poops (approved):

fastdivision (approved):

PR #195, opened 2021-08-04
Backup rates delete async

sethobey:
### Context

Users have experienced timeouts due to backup rate sync when a large number of nexus addresses are configured.

### Description

This PR builds on the previously implemented bulk action making rate creation asynchronous.
- Re-uses the existing chunk size already defined for async rate creation.
- Adds constants to the Taxjar Configuration class for message queue topics.
- Refactors many functionalities of the previous rates topic consumer to an abstract class, `AbstractRatesConsumer::class`, which the existing `Consumer::class` now extends from, as well as the newly created `DeleteRatesConsumer::class`
- Configures new topic for bulk action in message queue across various `etc/*.xml`

Screen Shot 2021-08-04 at 10 31 50 AM
_Messages displayed now upon clicking "sync rates"_

### Performance

For some users, the long-running synchronous task of deleting existing backup rates would lead to browser timeout. This change makes the process asynchronous, so no matter the number of backup rates stored, control of the browser will immediately be returned to the admin user while batched deletes process in the background.

### Testing

Prerequisites:
- TaxJar extension enabled in Magento 2
- At least one nexus address configured in TaxJar app and synced in Magento (preferably in a more tax-heavy US region such as California or Colorado)

1. From Admin dashboard, navigate to `Stores>Configuration>Sales>Tax` 
image
2. In dropdown box for "Backup Rates Enabled" select option "Yes" 
image
3. In top right, click "Save Config". image
4. Since there should not be any previous rates, no deletion should occur, but message should be displayed that bulk action to create backup rates has been queued.
5. Refresh the page and observe that backup rates are created
image
6. In the backup rates section, select "Sync Backup Rates" to force-refresh backup rates. 
image
7. Observe bulk action notifications that backup rates are being deleted as well as created. 
image
8. The outcome of delete/create can be verified by inspecting entries on the `tax_calculations`, `tax_calculation_rates` and `tax_calculation_rules` tables, as previous entries should no longer exist.

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Model/Import/AbstractRatesConsumer.php
+++ b/Model/Import/AbstractRatesConsumer.php
@@ -10,15 +10,13 @@ use Magento\Framework\Serialize\SerializerInterface;
 use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\EntityManager\EntityManager;
 use Taxjar\SalesTax\Model\Configuration as TaxjarConfig;
-use Taxjar\SalesTax\Model\Import\RateFactory;
-use Taxjar\SalesTax\Model\Import\RuleFactory;
 
-class Consumer
+abstract class AbstractRatesConsumer
 {
     /**
      * @var SerializerInterface
      */
-    private $serializer;
+    protected $serializer;
 
     /**
      * @var ScopeConfigInterface
@@ -33,7 +31,7 @@ class Consumer
     /**
      * @var TaxjarConfig
      */
-    private $taxjarConfig;
+    protected $taxjarConfig;
 
     /**
      * @var RateFactory
@@ -48,37 +46,42 @@ class Consumer
     /**
      * @var \Magento\AsynchronousOperations\Api\Data\OperationInterface
      */
-    private $operation;
+    protected $operation;
 
     /**
      * @var array
      */
-    private $rates;
+    protected $rates;
 
     /**
      * @var array
      */
-    private $newRates;
+    protected $newRates;
 
     /**
      * @var array
      */
-    private $newShippingRates;
+    protected $newShippingRates;
 
     /**
      * @var array
      */
-    private $productTaxClasses;
+    protected $productTaxClasses;
 
     /**
      * @var array
      */
-    private $customerTaxClasses;
+    protected $customerTaxClasses;
 
     /**
      * @var integer
      */
-    private $shippingClass;
+    protected $shippingClass;
+
+    /**
+     * @var string|null
+     */
+    protected $topicName;
 
     /**
      * @param SerializerInterface $serializer
@@ -95,8 +98,7 @@ class Consumer
         TaxjarConfig $taxjarConfig,
         RateFactory $rateFactory,
         RuleFactory $ruleFactory
-    )
-    {
+    ) {
         $this->serializer = $serializer;
         $this->scopeConfig = $scopeConfig;
         $this->entityManager = $entityManager;
@@ -117,6 +119,8 @@ class Consumer
         $this->reset();
 
         try {
+            $this->validate();
+            $this->setData();
             $this->processOperation();
         } catch (LocalizedException $e) {
             $this->fail($e, $e->getMessage());
@@ -134,20 +138,10 @@ class Consumer
     /**
      * Reset member variables as class is not reinstantiated between operations
      */
-    private function reset()
+    protected function reset()
     {
         $this->newRates = [];
         $this->newShippingRates = [];
-    }
-
-    /**
-     * @throws LocalizedException
-     */
-    private function processOperation()
-    {
-        $this->validate();
-        $this->setData();
-        $this->createRatesAndRules();
     }
 
     /**
@@ -155,7 +149,7 @@ class Consumer
      *
      * @throws LocalizedException
      */
-    private function validate()
+    protected function validate()
     {
         $this->validateBackupRatesAreEnabled();
         $this->validateAPIKeyIsSet();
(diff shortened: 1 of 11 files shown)


sethobey:
@dallendalton 
I had a few questions/notes from this implementation:

- Are UI displayed backup rates supposed to only “expect” taxable goods category? With multiple product tax categories selected, number of rate associations grows, but UI appears to only display taxable goods.
- This implementation is currently using the “deprecated” `delete()` method on `Rule`.. I would've preferred to use DI, but would like your guidance there as the RuleRepositoryInterface which I attempted to use was unable to locate the entry on `tax_calculation_rules` table
- Should I also replace the other usage of `ImportRates::_purgeRates()` and then remove method completely?
- There was no need to re-create TJ rule after deletion because the TJ tax calculation rule is re-created along with the incoming backup rates

sethobey:
@dallendalton - I've made (some of the) requested changes
- Abstract create and delete rates bulk operations into separate methods within `ImportRates.php`
- Made `AbstractRatesConsumer::setData` method abstract so that the concrete classes define their own implementation
- Renamed existing `Consumer.php` to `CreateRatesConsumer.php`

dallendalton:
@sethobey - Let's make sure we lint using the [Magento Code Standards](https://devdocs.magento.com/guides/v2.4/coding-standards/code-standard-php.html) and ensure everything passes.

Model/Import/AbstractRatesConsumer.php
```diff
@@ -134,28 +138,18 @@ public function process(\Magento\AsynchronousOperations\Api\Data\OperationInterf
...
-        $this->createRatesAndRules();
-    }
-
     /**
      * Validate that backup rates should be imported
      *
      * @throws LocalizedException
      */
-    private function validate()
+    protected function validate()
```
dallendalton: We may have to move this validate method to the create rates consumer. In the current version, when a customer disables the backup rates, or removes their connection to TaxJar, the delete rates consumer would never actually delete the rates due to this validation step and the rates would not be cleaned up.

Model/Import/AbstractRatesConsumer.php
```diff
@@ -117,6 +119,8 @@ public function process(\Magento\AsynchronousOperations\Api\Data\OperationInterf
         $this->reset();
 
         try {
+            $this->validate();
+            $this->setData();
```
dallendalton: The catch block for a generic exception below (line 130) needs updated to reference our TaxJar process rather then the product attributes update.

Observer/ImportRates.php
```diff
@@ -339,17 +324,19 @@ private function _importRates()
      *
      * @return OperationInterface
      */
-    private function makeOperation($rates, $bulkUuid): OperationInterface {
+    private function makeOperation($rates, $bulkUuid, $topic): OperationInterface
```
dallendalton: Let's abstract this method out a little further so that we aren't unnecessarily including the shipping_class, product_tax_classes, and customer tax classes parameters when making the delete rate operations. Perhaps a makeDeleteRatesOperation and makeCreateRatesOperation that calls a more generic make rates operation?

Observer/ImportRates.php
```diff
@@ -358,41 +345,84 @@ private function makeOperation($rates, $bulkUuid): OperationInterface {
...
-
-            try {
-                foreach ($calculations->getItems() as $calculation) {
-                    // @codingStandardsIgnoreStart
-                    $calculation->delete();
-                    // @codingStandardsIgnoreEnd
-                }
-            } catch (\Exception $e) {
-                $this->messageManager->addErrorMessage($e->getMessage());
+        if ($rule = $rateModel->getRule()) {
```
dallendalton: It appears that this if statement always evaluates to true, even if there is not a TaxJar rate stored in the database. It may be unnecessary.

Observer/ImportRates.php
```diff
@@ -358,41 +345,84 @@ private function makeOperation($rates, $bulkUuid): OperationInterface {
...
+                    TaxjarConfig::TAXJAR_TOPIC_DELETE_RATES
+                );
             }
 
-            try {
-                $this->rateRepository->deleteById($rate);
-            } catch (\Exception $e) {
-                $this->messageManager->addErrorMessage($e->getMessage());
+            if (! empty($operations)) {
+                $bulkDescription = __('Delete %1 TaxJar backup tax rates.', $this->batchSize);
```
dallendalton: As we discussed, I think it's unnecessary to include the batch size in the description.

Observer/ImportRates.php
```diff
@@ -358,41 +345,84 @@ private function makeOperation($rates, $bulkUuid): OperationInterface {
...
+        foreach ($rateChunks as $rateChunk) {
+            $operations[] = $this->makeOperation(
+                $rateChunk,
+                $bulkUuid,
+                TaxjarConfig::TAXJAR_TOPIC_CREATE_RATES
+            );
+        }
+
+        if (!empty($operations)) {
+            $bulkDescription = __('Create ' . $this->batchSize . ' TaxJar backup tax rates.');
```
dallendalton: Similar to with the delete rates, the batch size is probably unnecessary in the description.

sethobey:
Added unit test coverage for ImportRates observer

Screen Shot 2021-08-08 at 3 24 52 AM

dallendalton (approved):

PR #196, opened 2021-08-19
Int 1907 Correct test suite

sethobey:
### Context

The current Magento 2 extension's integration test suite has been broken and as such developers are unable to run the test suite during development. Without a functional test suite, we are unable to contribute test coverage to help prevent regression during refactor or feature development.

### Description

- Refactors `Transaction::class` to check for existence of an OrderItem before attempting to chain method call
- Adds test fixture "Builder" classes 
- Adds custom test cases for unit and integration tests
- Refactors integration `OrderTest::class` to use fixture builder.
- Refactors integration `RefundTest::class` to use fixture builder.
- Adds unit test coverage for `Refund::build()`

### Performance

Test suite did not run previously, and now it does.

### Testing

1. Install extension following instructions in README
2. Run `vendor/bin/phpunit --testsuite=Taxjar` 

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x


sethobey:
@dallendalton - Will be helpful in review to only diff the most recent commit.

**This PR is dependent on merge of #195**

I can rebase on develop branch instead if necessary.

Test/Integration/IntegrationTestCase.php
```diff
@@ -0,0 +1,38 @@
...
+     * @var CleanupReservationsInterface
+     */
+    protected $cleanupReservations;
+
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->objectManager = Bootstrap::getObjectManager();
+//        $this->cleanupReservations = $this->objectManager->get(CleanupReservationsInterface::class);
```
dallen-stripe, line 29: There is some commented out code here that needs to be cleaned up.

dallendalton: There is some commented out code here that needs cleaned up.

Test/BaseTestCase.php
```diff
@@ -0,0 +1,15 @@
...
+
+declare(strict_types=1);
+
+namespace Taxjar\SalesTax\Test;
+
+use PHPUnit\Framework\TestCase;
+
+class BaseTestCase extends TestCase
+{
+    public function todo(string $message = '')
```
dallendalton: I've been thinking about having this todo method in the BaseTestClass and I think its something we should talk about.

To me it feels similar to having a to do comment. i.e. //TODO:

I think an argument can be made for to do comments and for this method. However most of the time, it seems like eventually their original context or details get forgotten about but the to do comment remains. Future developers then may be hesitant to remove them as they know someone must have though they were important at some point. I think we would be better served by removing this and using our project management tools to track coverage we would like to add. What do you think?

Test/Integration/Model/Transaction/OrderTest.php
```diff
@@ -15,172 +15,128 @@
...
-        $this->cleanupReservations = Bootstrap::getObjectManager()->get(CleanupReservationsInterface::class);
-        $this->order = Bootstrap::getObjectManager()->get(Order::class);
-        $this->transactionOrder = Bootstrap::getObjectManager()->get('Taxjar\SalesTax\Model\Transaction\Order');
+        $this->taxjarOrder = $this->objectManager->get(TaxjarOrder::class);
     }
 
-    /**
-     * @magentoDataFixture ../../../../app/code/Taxjar/SalesTax/Test/Integration/_files/transaction/order_simple.php
-     */
     public function testDefaultOrder()
```
dallendalton, line 45: Perhaps we should rename some of these tests to better indicate what behavior we are testing. Maybe something like: testTotalsAreCorrect ? Though that doesn't quite fit since we are also testing the provider parameter here. Something to think about anyway.

dallendalton: Perhaps we should rename some of these tests to better indicate what behavior we are testing. Some of them are deceiving in what they actually test. Maybe for this method something like: testTotalsAreCorrect ? Though that doesn't quite fit since we are also testing the provider parameter here. Something to think about anyway.

Test/Integration/Model/Transaction/OrderTest.php
```diff
@@ -15,172 +15,128 @@
...
+        $result = $this->taxjarOrder->build($order);
 
-        $this->assertEquals(1, $result['customer_id'], 'Invalid customer ID');
+        $this->assertEquals($order->getCustomerId(), $result['customer_id'], 'Invalid customer ID');
     }
 
-    /**
-     * @magentoDataFixture ../../../../app/code/Taxjar/SalesTax/Test/Integration/_files/transaction/order_simple_ptc.php
-     */
     public function testExemptProductTaxClass()
```
dallendalton: This test no longer tests that the exempt product tax class is on the payload. It now checks that no tax class is set. Perhaps we should add another test to cover the original scenario.

Test/Integration/Model/Transaction/OrderTest.php
```diff
@@ -15,172 +15,128 @@
...
-     */
     public function testPositiveDecimals()
     {
-        $order = $this->order->loadByIncrementId('100000002');
-        $result = $this->transactionOrder->build($order);
+        $order = OrderBuilder::anOrder()->build();
+        $result = $this->taxjarOrder->build($order);
 
-        $this->assertEquals(27.0, $result['line_items'][0]['unit_price'], 'Invalid provider');
+        $this->assertEquals(10.0, $result['line_items'][0]['unit_price'], 'Invalid provider');
```
dallendalton: The message given if this test fails needs updated. 'Invalid provider' would be a misleading message if this test failed.

Test/Integration/Model/Transaction/OrderTest.php
```diff
@@ -15,172 +15,128 @@
...
+        $result = $this->taxjarOrder->build($order);
 
-        $this->assertEquals(5.0, $result['shipping'], 'Invalid shipping amount');
+        $this->assertEquals(20.0, $result['shipping'], 'Invalid shipping amount');
     }
 
-    /**
-     * @magentoDataFixture ../../../../app/code/Taxjar/SalesTax/Test/Integration/_files/transaction/order_simple_customer.php
-     */
     public function testExemptCustomer()
```
dallendalton: The name of this test is also deceptive as we are testing that the customer id on the payload is correct and not testing anything regarding the customers exemption status.

Test/Integration/Model/Transaction/RefundTest.php
```diff
@@ -1,183 +1,66 @@
...
+        $this->taxjarOrder = $this->objectManager->get(TaxjarOrder::class);
+        $this->taxjarRefund = $this->objectManager->get(TaxjarRefund::class);
     }
 
     /**
-     * @magentoDataFixture ../../../../app/code/Taxjar/SalesTax/Test/Integration/_files/transaction/refund_simple.php
+     * @magentoDbIsolation enabled
+     * @magentoAppIsolation enabled
      */
     public function testFullRefund()
```
dallendalton: Same thing as with some of the test names in the OrderTest class. I think the names of the methods could more clearly indicated the behavior that is being tested.

Test/Integration/_files/transaction/refund_bundle_partial.php
```diff
@@ -17,7 +17,7 @@
 
 require 'models/cleanup.php';
 require 'models/product_simple.php';
-require 'models/product_bundle.php';
+//require 'models/product_bundle.php';
```
dallendalton: Let's clean up this commented code here.

Util/Fixtures/Customer/AddressBuilder.php
```diff
@@ -0,0 +1,260 @@
...
+    public function buildWithoutSave(): AddressInterface
+    {
+        return clone $this->address;
+    }
+
+    private static function prepareFakeAddress(
+        ObjectManagerInterface $objectManager,
+        string $locale = 'en_US'
+    ): AddressInterface {
+        $faker = self::getAddressObject($locale);
```
dallendalton: Perhaps since we have removed faker as a dependency we should change the name of this variable. If I read this code I would assume the dependency was there. Maybe just fakeAddress would work?

Util/Fixtures/Catalog/ProductFixturePool.php
```diff
@@ -0,0 +1,47 @@
+<?php
```
dallendalton: Let's remove this class since it is no longer being used. Also the ProductFixtureRollback and ProductFixture classes.

dallendalton:
What would you think about moving the Util directory into the Test directory? I feel like doing so would fit a little more inline with normal Magento module structure since the files in the Util directory are only being used by the test suite.

Test/BaseTestCase.php
sethobey: I can get behind using `// TODO: ...` comments. I do think marking the test as skipped helps with visibility when devs are forced to observe yellow test suite pass instead of green, but I also agree that using the current project management tools should be the source of truth for tracking these issues and coverage gaps. As such, I'll remove this functionality.

Util/Fixtures/Customer/AddressBuilder.php
sethobey: Good catch

sethobey:
Okie dokie,  I think all should be resolved. Thanks so much for the thorough PR!

Here is my result from running tests after recommended changes:
image

dallendalton (approved):

PR #198, opened 2021-08-25
INT-1892: Backup rates progress indicator

sethobey:
### Context

The existing backup rates indicator shows how many rates exist, not specifically TaxJar backup rates. It is not a true indicator of how many of the available backup rates have been imported.

### Description

- Adds global configuration value for expected TaxJar backup rates count.
- Updates Backup Rates UI template (`backup.phtml`) and "controller" `Block.php` to provide data values from Block.php and render HTML in template (instead of concatenating a raw HTML string in Block.php)
- Removes Backup Rates UI element(s) for displaying individual states and their tax rate counts.
- Replaces "magic" strings with framework const values where applicable
- Updates cache flush method in `ImportRates.php`

Screen Shot 2021-08-24 at 11 13 03 AM

Screen Shot 2021-08-24 at 11 49 27 AM

### Performance

There is a somewhat negligible performance ding due to the cache clearing that is required to accurately display expected backup rates count when value is updated in global config.

### Testing

1. With TJ extension enabled and nexus synced in Magento 2, navigate to `Stores > Configuration > Sales > Tax`
2. Enable backup rates, then select "Save Config"
3. Observe values set for expected rates count and last updated at date.
4. Wait for async rates to process and observe that actual rates count matches expected rates count. (If verifying against DB values, note that any values that are not associated with TaxJar's `tax_calculation_rules` entry are not included in rate count, and the expected rates count value is set on table `core_config_data`.
5. Disable backup rates, then select "Save Config"
6. Wait for async rates delete process to finish, then observe TaxJar tax rule and related rates are deleted. Backup rates count value is set to `0` on table `core_config_data`.

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Observer/ImportRates.php
+++ b/Observer/ImportRates.php
@@ -20,20 +20,21 @@ declare(strict_types=1);
 namespace Taxjar\SalesTax\Observer;
 
 use Magento\AsynchronousOperations\Api\Data\OperationInterface;
+use Magento\AsynchronousOperations\Api\Data\OperationInterfaceFactory;
 use Magento\AsynchronousOperations\Model\Operation;
+use Magento\Authorization\Model\UserContextInterface;
 use Magento\Config\Model\ResourceModel\Config;
+use Magento\Framework\App\Cache\Manager as CacheManager;
 use Magento\Framework\App\Config\ScopeConfigInterface;
-use Magento\Framework\Event\ManagerInterface as EventManagerInterface;
-use Magento\Framework\Message\ManagerInterface as MessageManagerInterface;
+use Magento\Framework\Bulk\BulkManagementInterface;
+use Magento\Framework\DataObject\IdentityGeneratorInterface;
 use Magento\Framework\Event\Observer;
 use Magento\Framework\Event\ObserverInterface;
 use Magento\Framework\Exception\LocalizedException;
-use Magento\Tax\Model\Calculation\RateRepository;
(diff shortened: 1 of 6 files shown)


Observer/ImportRates.php
```diff
@@ -347,8 +341,6 @@ private function importRates(): void
         $this->messageManager->addSuccessMessage(
             __('TaxJar has successfully queued backup tax rate sync. Thanks for using TaxJar!')
         );
-
-        $this->eventManager->dispatch('taxjar_salestax_import_rates_after');
```
dallendalton: I'm not sure we want to remove this line. I know we don't use it anywhere in our extension but it may have been added as a way for other modules to run their functionality after we sync rates. It might be good to check back to where this line of code was added as see if there is any context for it.

sethobey: Ahh, I removed it because I didn't see it being used, but you may be right as it would be the only "after" hook/event, so I've re-added the line for now.

poops (approved):
LGTM!

dallendalton (approved):

PR #199, opened 2021-08-30
INT-1978: Backup rates upsert/delete

sethobey:
### Context

Acceptance testing of previous PR uncovered an error in rate creation/deletion.

Although the DB schema does not prescribe it, Tax Calculation Rates should be created with unique "codes" and the framework enforces this expectation. Because of this, and because of how TaxCalcRates were being created, if a TaxCalcRate already existed with the given code, it would not be updated, which would lead to inaccurate backup tax rates.

### Description

This PR corrects the issue described above by upsert-ing rates from the API request. For each rate from the request, if a Magento `tax_calculation_rate` already exists with the resulting code, that calculation rate will be updated from the request rate, otherwise a new calculation rate will be created from the request rate.

We also remove any calculation rates complementary to the request rates from the DB by creating a list of the resulting "codes" of all request rates, and then removing all Taxjar-specific calculation rates with codes that do not exist in the list. This modified delete method should work for all instances, as when the extension backup rates configuration is modified from enabled to disabled, a request will not be made, the resulting rates array will be empty and all existing TJ calculation rates would then be complementary.

As an added bonus, while testing, I found that with multiple product tax categories (PTCs) selected, the resulting count of rates displayed in the admin block would be the result of multiplying # of rates by # of PTCs, which turned out to be the number of `tax_calculation` entries created. For whatever reason, each calculation is viewed as an individual rate, even when multiple calculations relate to the same rate. A simple fix was added to the "Block' controller, using `array_unique()` to filter duplicate values when retrieving TJ-rule related rates.

### Performance

No noticeable change in performance. The creating of the request rate codes list may take some perceivable amount of time, but rate creation and deletion is still an asynchronous process, so control of browser session is quickly returned to admin user.

#### Recreate the problem:
1. Before checking out this branch, with backup rates enabled and synced
2. Modify the `rate` value of any Taxjar rate in the `tax_calculation_rates` table
3. From the store configuration page, run "Sync Rates"
4. Observe that the modified rate's value is not corrected/changed

#### Creation:
1. Enable backup rates and save configuration
2. Observe rates are created in bulk actions process
3. Number of rates created matches value stored in config on table `core_config_data`

#### Update with rate difference:
1. With backup rates enabled and synced
2. Modify some `tax_calculation_rates` entry's `rate` value to simulate when a request value differs from DB state
3. Observe the modified rate is now updated with expected value from request

#### Update with Nexus removal triggers deletion:
1. With backup rates enabled and synced, note the existing number of rates.
2. Visit `app.taxjar.com` and remove "nexus" from one region
3. In Magento 2, from the tax admin panel, select the option to sync backup rates.
4. Observe that the resulting rate count is now less that previously observed and any rates for the excluded nexus region are removed.

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Observer/ImportRates.php
+++ b/Observer/ImportRates.php
@@ -19,12 +19,14 @@ declare(strict_types=1);
 
 namespace Taxjar\SalesTax\Observer;
 
+use Magento\Framework\Bulk\OperationInterface as OperationInterfaceAlias;
 use Magento\Framework\Event\ManagerInterface as EventManagerInterface;
 use Magento\AsynchronousOperations\Api\Data\OperationInterface;
 use Magento\AsynchronousOperations\Api\Data\OperationInterfaceFactory;
 use Magento\AsynchronousOperations\Model\Operation;
 use Magento\Authorization\Model\UserContextInterface;
-use Magento\Config\Model\ResourceModel\Config;
+use Magento\Config\Model\ResourceModel\Config as ResourceConfig;
+use \Magento\Framework\App\Cache\Type\Config as CacheTypeConfig;
 use Magento\Framework\App\Cache\Manager as CacheManager;
 use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\Bulk\BulkManagementInterface;
@@ -36,10 +38,12 @@ use Magento\Framework\Message\ManagerInterface as MessageManagerInterface;
 use Magento\Framework\Serialize\SerializerInterface;
 use Magento\Tax\Model\Calculation\RateRepository;
 use Magento\Tax\Model\Config as MagentoTaxConfig;
+use Symfony\Component\HttpFoundation\Response;
 use Taxjar\SalesTax\Model\BackupRateOriginAddress;
 use Taxjar\SalesTax\Model\Client;
 use Taxjar\SalesTax\Model\ClientFactory;
 use Taxjar\SalesTax\Model\Configuration as TaxjarConfig;
+use Taxjar\SalesTax\Model\Import\Rate;
 use Taxjar\SalesTax\Model\Import\RateFactory;
 use Taxjar\SalesTax\Model\Import\RuleFactory;
 
@@ -66,7 +70,7 @@ class ImportRates implements ObserverInterface
     private $scopeConfig;
 
     /**
-     * @var Config
+     * @var ResourceConfig
      */
     private $resourceConfig;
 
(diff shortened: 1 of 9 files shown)


Observer/ImportRates.php
```diff
@@ -550,11 +573,12 @@ private function scheduleBulkOperation(array $data, string $topic, string $descr
...
      * @return array
      */
-    private function getRatesJson(): array
+    private function getRates(): array
     {
         $rates = $this->client->getResource('rates', [
-            '403' => __(
-                'Your last backup rate sync from TaxJar was too recent. Please wait at least 5 minutes and try again.'
+            Response::HTTP_FORBIDDEN => __(
+                'Your last backup rate sync from TaxJar was too recent. ' .
```
sethobey, line 584: Would we ever actually return a 403 here?

Observer/ImportRates.php
```diff
@@ -278,80 +281,134 @@ private function setZipCode(string $value): self
...
+                __('Debug mode enabled. Backup tax rates have not been altered.')
+            );
+            return;
+        }
+
+        $this->upsertRates($rates);
+        $this->purgeRates($codes);
+        $this->setRateCount($rateCount);
+        $this->setLastUpdate($updated);
+        $this->resourceConfig->saveConfig(TaxjarConfig::TAXJAR_BACKUP, (int) $isActive);
```
dallendalton, line 320: Why do we need to update this configuration here?

sethobey: In the "else" statement on the old version of this diff, this config value was set equal to false. Since the if/else was refactored away, I retained that functionality by including this line.. Although it was not apparent to me "why" we would want to do that.

It seems like the only way that it would really have any effect would be if backup rates were enabled, then the api key was removed or the extension was disabled.

I didn't see any harm in calling update each time, but if we only wanted to execute the line in the case mentioned above, we could wrap it in an if like so:
```
if (! $isActive) {
    $this->resourceConfig->saveConfig(TaxjarConfig::TAXJAR_BACKUP, 0);
}
```

dallendalton (approved):

poops (approved):

PR #203, opened 2021-09-22
INT-1953: Async backup rates test coverage

sethobey:
### Context

Missing test coverage for acceptance of Async Backup Tax Rates feature

### Description

Adds test coverage to help prevent regression
Identified all files that were modified in relation to this feature and added unit test coverage where practical.
For consumers, added basic integration tests.

### Performance

No impact

### Testing

1. Run Unit test configuration
2. Run integration test configuration

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x


Test/Mocks/ClientMock.php
```diff
@@ -0,0 +1,259 @@
...
+     * Perform a GET request
+     *
+     * @param string $resource
+     * @param array $errors
+     * @return array
+     */
+    public function getResource($resource, $errors = [])
+    {
+        return $this->mockResponse;
+//        $client = $this->getClient($this->_getApiUrl($resource));
```
dallendalton: Lets clean up this commented code here.

Test/Mocks/ClientMock.php
```diff
@@ -0,0 +1,259 @@
+<?php
```
dallendalton: Could this class be cleaned up a little? I think if this is just a mock then it would only need to implement the methods in the ClientInterface and would not have to implement all the private methods. I think also it could be simplified by having the other resource methods (postResource, etc) just return the mock response how your getResource method works.

Also, would it be more accurate to name/describe this class as a stub instead of a mock? Since it is not observing the internal behavior and is just providing test data?

Test/Mocks/ClientMock.php
```diff
@@ -0,0 +1,259 @@
...
+     * @return \Zend_Http_Client $client
+     */
+    private function getClient($url, $method = \Zend_Http_Client::GET)
+    {
+        // @codingStandardsIgnoreStart
+        $client = new \Zend_Http_Client($url, ['timeout' => 30]);
+        // @codingStandardsIgnoreEnd
+        $client->setUri($url);
+        $client->setMethod($method);
+//        $client->setConfig([
```
dallendalton: Theres some commented out code here that should probably be cleaned up. Also this is one of those private methods I mentioned in an earlier comment that I don't think are necessary for a mock class.

Test/Unit/Model/Import/CreateRatesConsumerTest.php
```diff
@@ -0,0 +1,284 @@
...
+            ->method('getApiKey')
+            ->willReturn('valid-api-key');
+
+        $sut = $this->getTestSubject();
+        $result = $this->callMethod($sut, 'validateApiKey');
+
+        self::assertInstanceOf(CreateRatesConsumer::class, $result);
+    }
+
+//    public function testPurgeStaleCalculations()
```
dallendalton: Some commented code that needs cleaned up here.

Test/Unit/Model/Import/CreateRatesConsumerTest.php
```diff
@@ -0,0 +1,284 @@
...
+        $this->scopeConfig = $this->createMock(ScopeConfigInterface::class);
+        $this->logger = $this->createMock(LoggerInterface::class);
+        $this->entityManager = $this->createMock(EntityManager::class);
+        $this->taxjarConfig = $this->createMock(TaxjarConfig::class);
+        $this->rateFactory = $this->getMockBuilder(RateFactory::class)->disableOriginalConstructor()->setMethods(['create'])->getMock();
+        $this->ruleFactory = $this->getMockBuilder(RuleFactory::class)->disableOriginalConstructor()->setMethods(['create'])->getMock();
+        $this->configCollection = $this->getMockBuilder(CollectionFactory::class)->disableOriginalConstructor()->setMethods(['create'])->getMock();
+    }
+
+    public function testSetCustomerTaxClasses()
```
dallendalton: For some of the test methods I'm not sure exactly what is being tested. It appears to me like the ones whose only assertion is assertInstanceOf are only testing that the sut can call a provide method and are not actually testing any expected outcomes. I'm not sure if they are really necessary.

In general I don't think private methods should be unit tested. Maybe an approach more like what you did in some of the other test classes would be of more value. Where you setup the mocks, call the public method and use the mock dependencies of the SUT to verify that the indented end result was achieved. Perhaps something like this:

$operationStub = $this->createMock(OperationInterface::class);
$operationStub->method('getSerializedData')->willReturn(whatever test data);
$rateMock->expects($this->once())->method('create')->with(whatever expected rate data)
$sut = $this->getTestSubject();
$sut->process($operationStub);

Test/Unit/Model/Import/RuleModelTest.php
```diff
@@ -0,0 +1,57 @@
...
+use Magento\Framework\Registry;
+use Magento\Tax\Model\Calculation;
+use Magento\Tax\Model\Calculation\Rule\Validator;
+use Magento\Tax\Model\ClassModel;
+use Taxjar\SalesTax\Model\Import\RuleModel;
+use Taxjar\SalesTax\Test\Unit\UnitTestCase;
+
+class RuleModelTest extends UnitTestCase
+{
+    public function testAfterSave()
```
dallendalton: Could a more descriptive name better indicated what result is being test. Maybe something like testAfterSaveDispatchesTaxSettingChangeAfter. That might be a little to verbose though?

Test/Unit/Model/Import/RuleModelTest.php
```diff
@@ -0,0 +1,57 @@
...
+                ['model_save_after'],
+                ['clean_cache_by_tags'],
+                ['tax_rule_save_after'],
+                ['tax_settings_change_after']
+            );
+
+        $mockContext = $this->createMock(Context::class);
+        $mockContext->expects($this->once())->method('getEventDispatcher')->willReturn($mockEventManager);
+
+        $sut = $this->getMockBuilder(RuleModel::class)
```
dallendalton, line 37: Do you need to mock the SUT here? Would it work to just instantiate the actual object with mock dependencies and then call the afterSave method?

Test/Unit/Model/Import/CreateRatesConsumerTest.php
sethobey: Hey @dallendalton - So I've renamed the most of the tests in this file to be a little more verbose.
 
Some of the first tests of setters like this one are smokey (and I can remove), but most of the tests that are asserting the return self are testing the "happy" path, and there is another test for when an exception is thrown.

This is one of the classes that we discussed would be a pain to test because all of the methods are private or protected and it doesn't have a public interface aside from the abstract class it extends from. The full functionality is validated in the Integration test for this class, but the kind of unit test described where we mock ALL of the dependency functionalities and pass them thru the constructor would end up with a lot of test setup that is super coupled to the code.

Maybe we can discuss this one in sync when you get a chance?

Test/Unit/Model/Import/RuleModelTest.php
sethobey: I don't really have to mock it here, and it would work by instantiating the object and passing the mock dependencies thru the constructor. The issue that I had with that is in the Tax module's Rule class constructor here:
image

The `_init` method in the tax module's class that our RateModel class extends from sets and retrieves the resource class, so we end up mocking deps that could be avoided by bypassing the constructor, but I can refactor this if you'd prefer.

dallendalton (approved):

poops (approved):

PR #213, opened 2021-10-19
INT-2041: Command flag option to force API transaction sync

sethobey:
### Context

Users have reached out regarding issues with syncing M2 transactions where the state of the order has changed, and those changes need to be reflected in the TaxJar service, but due to the comparison of a transaction's `updated_at `and `tj_salestax_last_sync` dates, the transaction will continue to be skipped.

Currently, the only acceptable workaround for this problem is for M2 administrators to manually modify date values in DB, which for a large number of transactions.

### Description

- Adds flag `-f|--force` to CLI command `taxjar:transaction:sync` image

- Refactors methods `\Taxjar\SalesTax\Model\Transaction\Order::push()` and `\Taxjar\SalesTax\Model\Transaction\Refund::push()` to accept optional parameters for http method and force-flag option.
- Adds "Force" select element (and tooltip) to the TaxJar transaction sync form 
image

- Adds unit test coverage for `\Taxjar\SalesTax\Model\Transaction\Backfill::process()` with `force` feature

### Performance

This should not have much if any affect on current extension performance,  but should resolve an on-going customer issue.

### Testing

1. With TaxJar's transaction sync feature enabled, *place an order*, either as customer, or as admin on behalf of customer, then invoice and mark shipped as necessary to move order status to "complete".
2. Verify that the order transaction has been persisted to TaxJar.
3. In DB, manually modify a value on the order such as an address value, or the order's `created_at` value.
4. From extension UI, attempt to sync transactions WITHOUT force-flag enabled. (using a date range containing the order)
5. Observe values not updated in TaxJar.
6. From extension UI, attempt to sync transactions WITH force-flag enabled (again, with proper date range).
7. Observe values updated in TaxJar.

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Block/Adminhtml/TransactionSync/Form.php
+++ b/Block/Adminhtml/TransactionSync/Form.php
@@ -104,12 +104,32 @@ class Form extends \Magento\Backend\Block\Widget\Form\Generic
         );
 
         $fieldset->addField(
-            'force',
-            'checkbox',
+            'force_sync_flag',
+            'select',
             [
-                'name' => 'force',
-                'label' => __('Force'),
-                'title' => __('Force'),
+                'name' => 'force_sync_flag',
+                'label' => __('Force sync'),
+                'title' => __('Enable force transaction sync'),
+                'values' => [
+                    [
+                        'value' => 0,
+                        'label' => 'No',
+                    ],
+                    [
+                        'value' => 1,
+                        'label' => 'Yes',
+                    ],
+                ],
+                'class' => 'select',
+                'after_element_html' => '
+                    <span class="tooltip-top" style="margin-left: 0.5rem">
+                        <a href="#" class="tooltip-toggle">What is this?</a>
+                        <span class="tooltip-content">
+                            The "Force sync" option will update all selected orders and credit
+                            memos in TaxJar regardless of last updated or synced dates.
+                        </span>
+                    </span>
+                ',
             ]
         );
 
--- a/Block/Adminhtml/TransactionSync/Form.php
+++ b/Block/Adminhtml/TransactionSync/Form.php
@@ -103,6 +103,16 @@ class Form extends \Magento\Backend\Block\Widget\Form\Generic
             ]
         );
 
+        $fieldset->addField(
+            'force',
+            'checkbox',
+            [
+                'name' => 'force',
+                'label' => __('Force'),
+                'title' => __('Force'),
+            ]
+        );
+
         $form->setAction('#');
         $form->setUseContainer(true);
         $form->setId(self::FORM_ELEMENT_ID);
--- a/Model/Transaction/Backfill.php
+++ b/Model/Transaction/Backfill.php
@@ -33,20 +33,22 @@ class Backfill
 {
     protected const OPERATION_DATA_KEY = 'meta_information';
 
+    protected const OPERATION_PARAMETER_KEYS = ['orderIds', 'force'];
+
     /**
      * @var OrderRepositoryInterface
      */
     protected $orderRepository;
 
     /**
-     * @var \Taxjar\SalesTax\Model\Transaction\OrderFactory
+     * @var Order
      */
-    protected $orderFactory;
+    private $orderTransaction;
 
     /**
-     * @var \Taxjar\SalesTax\Model\Transaction\RefundFactory
+     * @var Refund
      */
-    protected $refundFactory;
+    private $refundTransaction;
 
     /**
      * @var \Taxjar\SalesTax\Model\Logger
@@ -71,14 +73,6 @@ class Backfill
      * @var EntityManager
      */
     protected $entityManager;
-    /**
-     * @var Order
-     */
-    private $orderTransaction;
-    /**
-     * @var Refund
-     */
-    private $refundTransaction;
 
     /**
      * @param OrderRepositoryInterface $orderRepository
@@ -111,9 +105,8 @@ class Backfill
      */
     public function process(OperationInterface $operation): void
     {
-        $orderIds = $this->getOperationData($operation);
-
         try {
+            [$orderIds, $forced] = $this->getOperationData($operation);
             foreach ($orderIds as $orderId) {
                 $order = $this->orderRepository->get($orderId);
 
@@ -123,13 +116,13 @@ class Backfill
                 }
 
                 $this->orderTransaction->build($order);
-                $this->orderTransaction->push();
+                $this->orderTransaction->push($forced);
 
                 $creditMemos = $order->getCreditmemosCollection();
 
                 foreach ($creditMemos as $creditMemo) {
                     $this->refundTransaction->build($order, $creditMemo);
-                    $this->refundTransaction->push();
+                    $this->refundTransaction->push($forced);
                 }
             }
             $this->success($operation);
@@ -139,13 +132,32 @@ class Backfill
         }
     }
 
-    private function getOperationData($operation)
+    /**
+     * @param $operation
+     * @return array
+     * @throws LocalizedException
+     */
+    private function getOperationData($operation): array
     {
         $serialized = $operation->getSerializedData();
         $unserialized = $this->serializer->unserialize($serialized);
-        return $unserialized[self::OPERATION_DATA_KEY];
+        $data = $unserialized[self::OPERATION_DATA_KEY];
+        foreach (self::OPERATION_PARAMETER_KEYS as $key) {
+            if (!array_key_exists($key, $data)) {
+                throw new LocalizedException(
+                    __('Operation data could not be parsed. Required array key `%1` does not exist.', $key)
+                );
+            }
+        }
+        return array_map(function ($key) use ($data) {
+            return $data[$key];
+        }, self::OPERATION_PARAMETER_KEYS);
     }
 
+    /**
+     * @param $operation
+     * @throws \Exception
+     */
     private function success($operation)
     {
         $operation->setStatus(BulkOperationInterface::STATUS_TYPE_COMPLETE);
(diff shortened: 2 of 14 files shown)


Model/Transaction/Refund.php
```diff
@@ -115,70 +117,81 @@ public function build(
...
             }
+        }
+    }
 
-            // Retry push for existing records using PUT
-            if (!$forceMethod && $method == 'POST' && $error && $error->status == 422) {
-                $this->logger->log('Attempting to update refund / credit memo #' . $this->request['transaction_id'], 'retry');
-                return $this->push('PUT');
-            }
+    protected function makeRequest(string $method): array
```
dallendalton, line 179: It appears that this makeRequest method and the handleError method are almost identical between the refund and order models. Would it be worth it to abstract them out of these classes and into the transaction model?

Model/Transaction/Refund.php
```diff
@@ -115,70 +117,81 @@ public function build(
...
+        if ($method == Request::METHOD_PUT && $error->status == Response::HTTP_NOT_FOUND) {
+            $retry = Request::METHOD_POST;
+        }
+
+        if (isset($retry)) {
+            $this->logger->log(
+                sprintf('Attempting to retry saving refund / credit memo #%s', $this->request['transaction_id']),
+                'retry'
+            );
+            $this->push($retry);
```
dallendalton: I don't think this method call will work as intended now that the push method's signature has changed. I think it will need to be something like this:

$this->push($forced, $retry). 

We will probably need to carry through the forced flag into this method or make it a member variable. That way we still honor the original setting.

The same issue also exists in the handleError method on the order model.

Block/Adminhtml/TransactionSync/Form.php
```diff
@@ -103,6 +103,16 @@ protected function _prepareForm()
             ]
         );
 
+        $fieldset->addField(
```
dallendalton, line 106: Could we add a more details description to this field? I don't know if it will be clear to most customers what the functionality of the force option does.

Observer/BackfillTransactions.php
```diff
@@ -169,7 +173,10 @@ public function execute(Observer $observer): void
         $operations = [];
 
         foreach ($orderIdsChunks as $orderIdsChunk) {
-            $operations[] = $this->makeOperation($bulkUuid, $orderIdsChunk);
+            $operations[] = $this->makeOperation($bulkUuid, [
+                'orderIds' => $orderIdsChunk,
+                'force' => $observer->getData('force'),
```
dallendalton: When doing the backfill from the UI, the force field on the operation is always set to null. In my testing this occurred when the field is both checked and unchecked. This causes a TypeError to be thrown down the line as it is not a boolean.

dallendalton (changes requested):

dallendalton:
Also, it appears as if the logging isn't functioning properly. After running the backfill here is the log entries I am seeing for an order:

20 Oct 2021 22:45:44 - Pushing order #%s: %s
20 Oct 2021 22:45:44 - [API] Order #%s saved to TaxJar: %s

Where this is what it looks like currently (when using the develop branch):
23 Aug 2021 19:06:18 - [PUT] Pushing order #000000004: {"plugin":"magento","provider":"api","transaction_id":"000000004","transaction_date":"2021-08-23T19:00:35+0000","amount":100,"shipping":0,"sales_tax":7.25,"from_country":"US","from_zip":"84651","from_state":"UT","from_city":"Payson","from_street":"123 Main St","to_country":"US","to_zip":"84651","to_state":"UT","to_city":"payson","to_street":"123 main st","line_items":[{"id":"5","quantity":1,"product_identifier":"Configurable-Black","description":"Configurable","unit_price":100,"discount":0,"sales_tax":7.25,"product_tax_code":""}]}
23 Aug 2021 19:06:18 - [API] Order #000000004 updated in TaxJar: {"order":{"user_id":104920,"transaction_reference_id":null,"transaction_id":"000000004","transaction_date":"2021-08-23T19:00:35.000Z","to_zip":"84651","to_street":"123 main st","to_state":"UT","to_country":"US","to_city":"PAYSON","shipping":"0.0","sales_tax":"7.25","provider":"api","line_items":[{"unit_price":"100.0","sales_tax":"7.25","quantity":1,"product_tax_code":null,"product_identifier":"Configurable-Black","id":0,"discount":"0.0","description":"Configurable"}],"from_zip":"84651","from_street":"123 Main St","from_state":"UT","from_country":"US","from_city":"PAYSON","exemption_type":null,"customer_id":null,"amount":"100.0"}}

Model/Transaction/Refund.php
sethobey: Wow, great catch. I made a stupid git error right before I pushed this PR.  I thought that I had recovered everything properly in the IDE, but obviously not..

Model/Transaction/Refund.php
sethobey: I agree with that as well, which is why I'm doing that in INT-1942 currently. I wanted to wait for that ticket since it seemed like it would touch these classes more.

Block/Adminhtml/TransactionSync/Form.php
sethobey: Sure thing, I think thats a good call out

sethobey: So in response to your feedback, I decided on two things:
1. To address the primary issue of ambiguity, I've added a hoverable tooltip following the form field in the adminhtml UI that describes the feature.
2. I think it kind of touches on ambiguity a bit, but also just provides a safer UI, so I changed the checkbox to a select element instead with declarative options "No" and "Yes". I think it is more inline with current UI than my original decision and might help prevent some situations where a customer accidentally engages with the form element esp. on touch devices like mobile.

Screen Shot 2021-10-21 at 7 11 28 PM

Observer/BackfillTransactions.php
sethobey: ✅

sethobey:
And log message string template errors are corrected also

[transaction.log.txt](https://github.com/taxjar/taxjar-magento2-extension/files/7393580/transaction.log.txt)

Model/Transaction/Backfill.php
```diff
@@ -111,25 +105,24 @@ public function __construct(
      */
     public function process(OperationInterface $operation): void
     {
-        $orderIds = $this->getOperationData($operation);
-
         try {
+            [$orderIds, $forceFlag] = $this->getOperationData($operation);
```
dallendalton, line 109: There is still an issue with the force flag. Now I correctly see that the force flag is being set to 0 when the operation is created in BackfillTransactions.php. However here when it is getting the deserialized data from the operation is getting the string "0". Which causes a type error in the order push method as it is expecting a boolean.

dallendalton (dismissed):

Model/Transaction/Order.php
```diff
@@ -108,26 +106,32 @@ public function push(string $method = null) {
...
+            $this->logger->log(
+                sprintf(
+                    'Order #%s saved to TaxJar: %s',
+                    $this->request['transaction_id'],
+                    json_encode($response)
+                ),
+                'api'
+            );
 
             $this->originalOrder->setData('tj_salestax_sync_date', gmdate('Y-m-d H:i:s'))->save();
```
dallendalton: When running transaction backfill, after the order is synced it updates the sync timestamp. This triggers our extension to immediately try and sync it again. While our validation around updated at and sync at stops the second sync from triggering and API request it still clutters the logs and is not ideal. I wonder if there is a way we can still update the sync timestamp but not trigger the event/observer that causes the synchronous sync.

Model/Transaction/Backfill.php
```diff
@@ -111,25 +105,24 @@ public function __construct(
...
+                if (!$this->orderTransaction->isSyncable($order)) {
                     $this->logger->log('Order #' . $orderId . ' is not syncable', 'skip');
                     continue;
                 }
 
                 $this->orderTransaction->build($order);
-                $this->orderTransaction->push();
+                $this->orderTransaction->push($forceFlag);
 
                 $creditMemos = $order->getCreditmemosCollection();
```
dallendalton, line 121: This line appears not be working as intended. I tested by triggering the transaction sync through the UI without the force option enabled. 

When it gets to this line it doesn't find any creditMemos for the order. I tested on the develop branch using the same date range and it correctly found all the creditMemos for the orders selected. I'm not really sure what is causing the difference though.

dallendalton (changes requested):

sethobey:
> And log message string template errors are corrected also
> 
> [transaction.log.txt](https://github.com/taxjar/taxjar-magento2-extension/files/7393580/transaction.log.txt)

@dallendalton Taking a look in this log now, I see those `[SKIP]` entries hiding in there from the order-save event re-triggering

Model/Transaction/Order.php
sethobey: This is resolved by refactoring to only save only the attribute, instead of the model (like in Refund class), bypassing the model saved event

Model/Transaction/Refund.php
```diff
@@ -116,69 +118,103 @@ public function build(
...
+            default:
+                throw new LocalizedException(
+                    __('Unhandled HTTP method "%s" in TaxJar refund transaction sync.', $method)
+                );
+        }
+    }
+
+    /**
+     * Handles case where error occurs when POSTing a resource that already
+     * exists of PUTing a resource that has not been created yet.
```
dallendalton: Minor typo in the comment here, should be or instead of of I think.

dallendalton (approved):
One minor typo, but other than that looks great.

Model/Transaction/Refund.php
sethobey: 🦅 👀

jordan-k-johnson (approved):
LGTM

PR #214, opened 2021-10-20
INT 1963: Sync partially cancelled order items

sethobey:
### Context

M2 extension does not account for "cancelled" items on completed orders.

When an order is created for some positive quantity (X), a quantity less than X is invoiced and shipped, and the remaining quantity is subsequently cancelled, no credit memo (refund) is created for the cancelled items since they were never invoiced. Since there is not credit memo, the API extension counts the cancelled item quantities as valid and still parses them to be sent to TaxJar.

### Description

The act of invoicing an order creates the taxable event, therefore it should actually be the _invoiced_ amount that is used in line item calculations. The tax burden is dependent upon some exchange of goods, so that the act of "creating" the order in M2 is more akin to creating a COD purchase order request, and it is not until invoicing that a taxable exchange has occurred.

For those reasons, the `\Taxjar\SalesTax\Model\Transaction\Order::class` class has been refactored to use the actual invoiced amounts when building transaction sync request payloads. The `\Taxjar\SalesTax\Model\Transaction::class` class now uses the invoiced quantity instead of the originally requested quantity.

### Performance

Corrects known issue, but no measurable performance gains or losses.

### Testing

1. Checkout the "develop" main branch.
2. With TaxJar activated and transaction sync enabled, place an order on behalf of customer for qty > 1.
3. Invoice the order for less than total quantity and ship that same less-than-total quantity, then cancel the remaining portion of the order.
4. Observe that the order's "grand total" and "tax total" does not reflect what is shown in M2 as the cost of the cancelled item(s) is/are still included in the TaxJar transaction amounts.
5. Checkout this PR branch, run setup DI and upgrade commands.
6. Sync backup rates with "force" flag either via UI or CLI.
7. Observe that values in TaxJar have been overwritten with updated values which account for cancelled item and tax amounts.

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -19,7 +19,6 @@ namespace Taxjar\SalesTax\Model;
 
 use Magento\Bundle\Model\Product\Price;
 use Magento\Framework\Exception\LocalizedException;
-use Magento\Framework\Exception\NoSuchEntityException;
 use Magento\Sales\Api\Data\OrderInterface;
 use Taxjar\SalesTax\Helper\Data as TaxjarHelper;
 use Taxjar\SalesTax\Model\Configuration as TaxjarConfig;
@@ -250,8 +249,13 @@ class Transaction
             }
 
             $itemId = $item->getOrderItemId() ? $item->getOrderItemId() : $item->getItemId();
-            $discount = (float) $item->getDiscountAmount();
-            $tax = (float) $item->getTaxAmount();
+            $discount = (float) $item->getDiscountInvoiced();
+            $tax = (float) $item->getTaxInvoiced();
+
+            if ($type == 'refund' && isset($creditMemoItem)) {
+                $discount = (float) $creditMemoItem->getDiscountAmount();
+                $tax = (float) $creditMemoItem->getTaxAmount();
+            }
 
             if (isset($parentDiscounts[$itemId])) {
                 $discount = $parentDiscounts[$itemId] ?: $discount;
--- a/Model/Transaction.php
+++ b/Model/Transaction.php
@@ -219,7 +219,7 @@ class Transaction
 
             $parentItem = $item->getParentItem();
             $unitPrice = (float) $item->getPrice();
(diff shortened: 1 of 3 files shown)


Model/Transaction.php
```diff
@@ -219,7 +219,7 @@ protected function buildLineItems($order, $items, $type = 'order'): array
 
             $parentItem = $item->getParentItem();
             $unitPrice = (float) $item->getPrice();
-            $quantity = (int) $item->getQtyOrdered();
+            $quantity = (int) $item->getQtyInvoiced();
 
             if ($type == 'refund' && isset($creditMemoItem)) {
```
dallendalton, line 223: It looks like there is still an issue. When testing the line items in the request still had the tax amount from the original order. So for example quantity 2 of a $100 item with a 10% tax rate. One item is cancelled so the line item should look like this:

{
unit_price:100,
sales_tax:10,
quantity:1,
product_tax_code:"",
product_identifier:"Simple",
id:"13",
discount:0,
description:"Simple"
}

But actually looks like this:

{
unit_price:100,
sales_tax:20,
quantity:1,
product_tax_code:"",
product_identifier:"Simple",
id:"13",
discount:0,
description:"Simple"
}

dallendalton (changes requested):

Model/Transaction.php
sethobey: Yeah, I was mistaken and the `$item` variable listed is always the `Order\Item::class` and never the `CreditMemo\Item::class`. I added a check that if type is equal to refund, then use the tax and discount amounts from the creditmemo and it looks correct to me now.

jordan-k-johnson (approved):

dallendalton (approved):

PR #218, opened 2021-11-10
INT 2085: Optimize transaction backfill order search

sethobey:
### Context

Current `BackfillTransactions:class` observer uses SearchCriteria to retrieve orders within given date and store constraints, but does not consider whether the order is actually "syncable" before batching that order into a bulk operation. The consumer `Backfill::class` then determines whether or not we should sync the transaction. This means that potentially many unnecessary operations are created only to determine that the orders within should not be synced.

### Description

This PR will now check within `BackfillTransactions:class` whether or not "force" option was enabled, and if force is not enabled, only syncable orders will be included in operations. While adding this functionality, `BackfillTransactions.php` was refactored, unit and integration tested.

_Also corrects failing Transaction::class unit tests_

### Performance

The same total number of order-date comparisons will now be paid upfront instead spread across consumer processes, but there should be a net decrease in number of operations for the same request.

#### Automated Testing
-  Adds `Taxjar\SalesTax\Test\Unit\Observer\BackfillTransactionsTest.php`
-  Adds `Taxjar\SalesTax\Test\Integration\Observer\BackfillTransactionsTest.php`

#### Manual Testing
1. Enable Taxjar transaction sync
2. Create/invoice/ship an order (optionally create credit memo as well)
3. Observe order is synced with Taxjar
4. Trigger backfill transaction sync (_not forced_) with date range inclusive of order
5. Observe created order not included in any created operations
6. Trigger _forced_ backfill transaction sync with date range inclusive of order
7. Observe created order is now included in created operations

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x


Observer/BackfillTransactions.php
```diff
@@ -15,309 +15,321 @@
...
-            );
+        try {
+            if (!$this->taxjarConfig->getApiKey()) {
+                throw new \Magento\Framework\Exception\LocalizedException(
+                    __('Could not sync transactions with TaxJar. Please make sure you have an API key.')
+                );
+            }
+
+            $criteria = $this->getSearchCriteria();
+            $orders = $this->getOrders($criteria, $this->getInput('force'));
```
dallendalton: When running a backfill from the UI both with force enabled and without I am getting the following error:

Argument 2 passed to Taxjar\SalesTax\Observer\BackfillTransactions::getOrders() must be of the type bool or null, string given, called in /var/www/html/app/code/Taxjar/SalesTax/Observer/BackfillTransactions.php on line 135

It looks like $this->getInput('force') is returning the string '0' when force is not enabled. I would guess it is returning '1' when it is enabled.

dallendalton (changes requested):
When running a backfill without force enab

Observer/BackfillTransactions.php
sethobey: Ahh crap.. nice. Think I might've added that bool type while writing tests.

sethobey: Added casts to bool and updated unit tests to account for this scenario

jordan-k-johnson (approved):

Observer/BackfillTransactions.php
```diff
@@ -15,309 +15,321 @@
...
+     * @throws \Magento\Framework\Exception\LocalizedException
+     */
+    public function success()
+    {
+        $message = 'Transaction sync successfully scheduled. ';
+        $message .= json_encode($this->getConfiguration());
 
-        $criteria = $this->getSearchCriteria($observer->getData());
-        $orderResult = $this->orderRepository->getList($criteria);
+        $this->logger->log($message);
```
dallendalton: Let's separate the log message from what get's displayed in the UI. In the UI I think the successful message should be something like:

"Successfully scheduled 5 transactions to be synced to TaxJar."

Then for the log we could included that message but then have the additional configuration details. Something like this perhaps:

"Successfully scheduled 5 transactions to be synced to TaxJar. Details: {"from":sometimestamp,"to":sometimestamp,"force_sync":false}"

In the failure case we should do them same. Give the failure message and the exemption message back to the UI. Then for the log entry give those messages plus the configuration data.

Observer/BackfillTransactions.php
```diff
@@ -15,309 +15,321 @@
...
-        ];
-        $data = [
-            'data' => [
-                'bulk_uuid' => $bulkUuid,
-                'topic_name' => TaxjarConfig::TAXJAR_TOPIC_SYNC_TRANSACTIONS,
-                'serialized_data' => $this->serializer->serialize($dataToEncode),
-                'status' => \Magento\Framework\Bulk\OperationInterface::STATUS_TYPE_OPEN,
-            ]
+    private function getConfiguration(): array
+    {
```
dallendalton, line 363: When triggering a sync from the UI I am getting this:

image

It appears it is not correctly getting the to and from parameters.

Update: on further investigation it looks like this is an issue with the get Input function.

dallendalton (changes requested):

Observer/BackfillTransactions.php
sethobey: @dallendalton - Did you select "to" and "from" dates? I saw the same output when I didn't select any dates, but that was expected because the JSON is the user/system input, not the resulting search criteria used for the order lookup.

If you did select dates from the input fields and that was the output, then that is unexpected. I'll try it again on my local.

dallendalton: @sethobey I selected the inputs. If no inputs are selected though we should still display the timeframe that was searched and not a null value. That way the logs are clear on exactly what search was performed.

Observer/BackfillTransactions.php
sethobey: I agree with the premise, but this one is easier said than done, because what is shown in the UI is the `playback()` of console output for our Logger class, kind of just "using" a side-effect of how `Logger::class` works.  To properly handle this situation, I would need to either change functionality of `Logger::class`, which might be out of scope, or persist only the part of the message we want shown in some retrievable way from the frontend.

You can see how the response we display in UI is generated here:
image

If we wanted to continue using the Logger class, this could probably be accomplished pretty easily with either new method(s) on the Logger class, or refactor what is already there, but might also be worth more discussion since we are already planning other changes to logging.

Observer/BackfillTransactions.php
sethobey: @dallendalton - This should be resolved now. I had renamed the expected AJAX param in PHP and updated tests accordingly, but missed it on the frontend.

Observer/BackfillTransactions.php
sethobey: Resolved with addition of `playback` setter-method on Logger class and refactor of BackfillTransactions success and fail methods.

Model/Logger.php
```diff
@@ -119,29 +132,32 @@ public function force($isForced = true)
...
      * Save a message to taxjar.log
      *
      * @param string $message
      * @param string $label
-     * @throws LocalizedException
+     * @throws \Magento\Framework\Exception\LocalizedException
      * @return void
      */
-    public function log($message, $label = '')
+    public function log(string $message, string $label = ''): void
```
dallendalton: Changing the signature of this method appears to have broken some other portions of the integration. I am getting this error:

Fatal error: Uncaught TypeError: Argument 2 passed to Taxjar\SalesTax\Model\Logger::log() must be of the type string, null given, called in /var/www/html/app/code/Taxjar/SalesTax/Model/Transaction/Refund.php on line 150 and defined in /var/www/html/app/code/Taxjar/SalesTax/Model/Logger.php:154
Stack trace:
#0 /var/www/html/app/code/Taxjar/SalesTax/Model/Transaction/Refund.php(150): Taxjar\SalesTax\Model\Logger->log('Pushing refund ...', NULL)
#1 /var/www/html/app/code/Taxjar/SalesTax/Model/Transaction/Backfill.php(125): Taxjar\SalesTax\Model\Transaction\Refund->push(true)
#2 [internal function]: Taxjar\SalesTax\Model\Transaction\Backfill->process(Object(Magento\AsynchronousOperations\Model\Operation))
#3 /var/www/html/vendor/magento/framework-message-queue/Consumer.php(150): call_user_func(Array, Object(Magento\AsynchronousOperations\Model\Operation))
#4 /var/www/html/vendor/magento/framework-message-queue/Consumer.php(218): Magento\Framework\MessageQueue\Consumer->dispatchMessage(Object(Magento\Framework\MessageQueue\Env in /var/www/html/app/code/Taxjar/SalesTax/Model/Logger.php on line 154

This was when attempting a force backfill that found a single order with a single credit memo. It looks like null is being pass as the second parameter of the log command.

Observer/BackfillTransactions.php
```diff
@@ -15,309 +15,347 @@
...
         }
+    }
 
-        $this->logger->log(__('Initializing TaxJar transaction sync'));
+    /**
+     * @throws \Magento\Framework\Exception\LocalizedException
+     */
+    public function success()
+    {
+        $this->log('Transaction sync successfully scheduled.');
```
dallendalton: What would you think about including the number of transactions we scheduled to sync. So for example say we found 5 orders it would be something like:

"Successfully scheduled transaction sync for 5 orders and their credit memos".

Then if we don't find any orders that need to be synced it could be something like:

"No orders were found that need to be synced to TaxJar".

The messaging might need a little work but in general I think it might make it more clear to the merchant what is happening.

Observer/BackfillTransactions.php
```diff
@@ -15,309 +15,347 @@
...
+            $criteria = $this->getSearchCriteria()->__toArray();
+        } catch (\Exception $e) {
+            $criteria = null;
+        }
+
+        return [
+            'from' => $this->getInput('from'),
+            'to' => $this->getInput('to'),
+            'force_sync' => (bool)$this->getInput('force'),
+            'criteria' => $criteria,
```
dallendalton: I'm not sure if we need to include the criteria in the logged configuration data. Because we generate the criteria only using the from and to inputs those may be sufficient to replicate any issue that occurred by using the logs.

dallendalton (changes requested):

dallendalton (approved):

Issue #194, opened 2021-07-22
Address validation fails to save phone number when suggesting validated version of address already saved to customer


bm-eventstable:
### Prerequisites

- [x] Are you running the latest version of our extension?
- [x] Have you reviewed our [integration guide](https://www.taxjar.com/guides/integrations/magento2/)?
- [x] Did you search through [existing issues](https://github.com/taxjar/taxjar-magento2-extension/issues)?

### Description
During checkout as a registered customer, if an existing saved address is selected and the Taxjar module suggests a validated address. If the customer chooses to use the Taxjar validated address, they will not be able to complete checkout as the phone number is not copied to the updated (TaxJar suggested) address.

### Steps to Reproduce

1. Begin checkout ass suggested customer
2. Select shipping address from one saved to customer record
3. Now select to use Taxjar validated (corrected) address
4. Proceed to the payment step, enter payment details
5. Attempt to complete checkout

### Expected Result
Checkout is successful

### Actual Result
Error message received that phone number is required for the shipping address (and billing address if set to use the same for both).

### Versions
This occurred while running Magento Commerce Cloud 2.3.7 with PHP 7.3 and version 1.6.5 of your TaxJar module

lagunacellar:
We ran into the same issue. How come Taxjar has not released a fix 4 months after the initial report?

sethobey:
Thank you @bm-eventstable and @lagunacellar, this issue has been prioritized and a fix should be included in our next patch release.

beechyrb:
@sethobey When will this be included in a release? We started having these issues a few days ago and it is really frustrating for our customers. I will apply the patch for now.

sethobey:
Hi @beechyrb, soon! This fix will be included in our upcoming v1.8.0 release which is currently pending review for Magento marketplace release.

sethobey:
CC @beechyrb, @bm-eventstable, @lagunacellar 

As of this morning, v1.8.0 is available on Marketplace and Packagist. Thanks everyone!

PR #219, opened 2021-11-11
issue #194: phone number is required for the shipping address

sanchit-redmonks:
### Context
Patch for issue #194


sethobey:
Hello @sanchit-redmonks, Issue #194 is already assigned internally, but thank you for your PR submission!

PR #222, opened 2021-11-22
INT-2121: Sales order view transaction sync

sethobey:
### Context

Other TJ integrations have the ability to sync a specific order from UI, but integration would currently require an admin to determine the date that an order was originally placed, then initiate transaction backfill (with force sync enabled) over a date range containing the order's date.

### Description

Adds a "Sync to TaxJar" button to the UI toolbar on the Magento Sales Order view page when an order is complete. Clicking the button will force an API request to sync that individual order and any related credit memos with TaxJar.

### Performance

The addition of this feature should improve user experience while increasing feature parody across TJ connectors.

### Testing

- Manually verified "Sync to TaxJar" button is only present on completed orders
- Unit test coverage of added TJ helper methods
- Unit test coverage of controller classes

1. Enable TJ module and "transaction sync" feature.
2. Create an order
3. View order, observe "Sync to TaxJar" button not displayed, Order not synced.
4. Invoice the order
5. Observe "Sync to TaxJar" button not displayed, Order not synced.
6. Ship the order
7. Observe "Sync to TaxJar" button is now displayed, Order is completed and synced to TaxJar.
8. Click "Sync to TaxJar"
9. Observe sales tax synced at date time is updated to "now", API call can be verified by viewing  `taxjar/transaction.log`

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3
- [ ] Magento 2.2
- [ ] Magento 2.1

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x
- [ ] PHP 5.x

### DIFF
--- a/Plugin/Sales/Block/Adminhtml/Order/View.php
+++ b/Plugin/Sales/Block/Adminhtml/Order/View.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Taxjar_SalesTax
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Taxjar
+ * @package    Taxjar_SalesTax
+ * @copyright  Copyright (c) 2020 TaxJar. TaxJar is a trademark of TPS Unlimited, Inc. (http://www.taxjar.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace Taxjar\SalesTax\Plugin\Sales\Block\Adminhtml\Order;
+
+class View
+{
+    /**
+     * @var \Taxjar\SalesTax\Helper\Data
+     */
+    protected $_tjSalesTaxData;
+
+    /**
+     * @param \Taxjar\SalesTax\Helper\Data $tjSalesTaxData
+     */
+    public function __construct(\Taxjar\SalesTax\Helper\Data $tjSalesTaxData)
+    {
+        $this->_tjSalesTaxData = $tjSalesTaxData;
+    }
+
+    /**
+     * @param \Magento\Sales\Block\Adminhtml\Order\View $subject
+     */
+    public function beforeSetLayout(\Magento\Sales\Block\Adminhtml\Order\View $subject)
+    {
+        if ($this->_tjSalesTaxData->isSyncableOrder($subject->getOrder())) {
+            $subject->addButton('taxjar_sync', [
+                'label' => __('Sync to TaxJar'),
+                'class' => 'taxjar-sync primary',
+                'onclick' => 'syncTransaction(\'' . $subject->getOrderId() . '\')'
+            ]);
+        }
+    }
+}
+
--- a/Plugin/Sales/Block/Adminhtml/Order/View.php
+++ b/Plugin/Sales/Block/Adminhtml/Order/View.php
@@ -39,13 +39,23 @@ class View
      */
     public function beforeSetLayout(\Magento\Sales\Block\Adminhtml\Order\View $subject)
     {
-        if ($this->_tjSalesTaxData->isSyncableOrder($subject->getOrder())) {
+        if ($this->shouldRender($subject->getOrder())) {
             $subject->addButton('taxjar_sync', [
                 'label' => __('Sync to TaxJar'),
                 'class' => 'taxjar-sync primary',
                 'onclick' => 'syncTransaction(\'' . $subject->getOrderId() . '\')'
             ]);
         }
+    }
+
+    /**
+     * @param \Magento\Sales\Api\Data\OrderInterface $order
+     * @return bool
+     */
+    protected function shouldRender(\Magento\Sales\Api\Data\OrderInterface $order): bool
+    {
+        return $this->_tjSalesTaxData->isTransactionSyncEnabled()
+            && $this->_tjSalesTaxData->isSyncableOrder($order);
     }
 }
 
--- a/view/adminhtml/templates/order/view/synced.phtml
+++ b/view/adminhtml/templates/order/view/synced.phtml
@@ -15,24 +15,47 @@
  * @license    http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
  */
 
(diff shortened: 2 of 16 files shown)


Plugin/Sales/Block/Adminhtml/Order/View.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->_tjSalesTaxData = $tjSalesTaxData;
+    }
+
+    /**
+     * @param \Magento\Sales\Block\Adminhtml\Order\View $subject
+     */
+    public function beforeSetLayout(\Magento\Sales\Block\Adminhtml\Order\View $subject)
+    {
+        if ($this->_tjSalesTaxData->isSyncableOrder($subject->getOrder())) {
```
dallendalton: It appears that this allows the sync button to be displayed even when transaction sync is disabled. This creates a scenario where the sync actually fails (but without throwing an exception) and results in this alert box:

image

When transaction sync is disable in the scope of the order I believe that the button should not be displayed.

view/adminhtml/templates/order/view/synced.phtml
```diff
@@ -15,24 +15,47 @@
...
+                 url: '<?= $block->escapeUrl($block->getUrl('taxjar/transaction/sync')) ?>',
+                 method: 'POST',
+                 data: {
+                     order_id: orderId
+                 }
+             }).done(function(data) {
+                 if (data.result) {
+                     window.location.reload();
+                 } else {
+                     alert('Could not sync to TaxJar! Error: ' + data.error_message);
```
dallendalton: I'm not sure that alerting here is the best way to handle errors. We already pass the exception message using the message manager ($this->messageManager->addErrorMessage).

Using alert (rather than just reloading the page regardless of result), can result in a scenario where the user can click sync multiple time. Each time it fails it adds a message to the message manager which can lead to this displaying when they reload the page (or even navigate to another page).

image

Just reloading the page would also handle the scenarios where an the transaction is not synced but an exception is not thrown.

view/adminhtml/templates/order/view/synced.phtml
```diff
@@ -15,24 +15,47 @@
...
+             syncButton.find('span').text('Syncing...');
+
+             $.ajax({
+                 url: '<?= $block->escapeUrl($block->getUrl('taxjar/transaction/sync')) ?>',
+                 method: 'POST',
+                 data: {
+                     order_id: orderId
+                 }
+             }).done(function(data) {
+                 if (data.result) {
```
dallendalton: data.result is never set, so even on a successful sync it show the error alert box. I believe this was intended to be data.success.

dallendalton (changes requested):
I found a few issues and left comments on the corresponding code locations.

view/adminhtml/templates/order/view/synced.phtml
sethobey: This was me trying to be clever, but in retrospect, I'm really not sure why I didn't just reload page regardless of request outcome.. lol

view/adminhtml/templates/order/view/synced.phtml
sethobey: > `data.result` was being set, but only set on "successful" requests.

Actually.. no its not, my bad. 💀

Plugin/Sales/Block/Adminhtml/Order/View.php
sethobey: This is a good find and I have fixed it, but while addressing it uncovered another scenario with similar logic to consider..

Do we want to display the TJ transaction sync date in the Sales Order info UI even if transaction sync is disabled?

Currently we do, but is that necessarily a good thing? I think that unlike this bug you've found here, where we obviously only want to show the "Sync to Taxjar" button (or include related JS) if transaction sync is enabled, it makes more sense if this sync-date partial is always included _if TaxJar is enabled_.

Take the case where transaction sync feature was enabled, but user no longer wishes to use it and disables it. That TJ sync date will still be displayed (i.e. the `.phtml` template/partial will be applied to the page) if the order has a `tj_salestax_sync_date` value stored on it, regardless of whether that feature is enabled. The argument could be made that because the data exists, it should be available to admin, but I'm not sure I agree.

If after installing the TJ SalesTax module, you decide to uninstall it, clear static content, recompile, and run upgrade, the `tj_salestax_sync_date` column will still exist on the `sales_order` table, and any existing values will remain. That data is still there, just hidden because TJ is "turned off" in M2. While that is partly due to the fact that our module doesn't include extension attributes properly for legacy reasons, I think it is also indicative of the fact that whether or not we included this UI template should depend on if our module is enabled, not the state of an order. The state of the order will then affect what we display within that UI template.

sethobey: This is a good find and I have fixed it, but while addressing it uncovered another scenario with similar logic to consider..

Do we want to display the TJ transaction sync date in the Sales Order info UI even if transaction sync is disabled?

Currently we do, but is that necessarily a good thing? I think unlike this(^) bug you've found here, where obviously we only want to show the "Sync to Taxjar" button (or include related JS) if transaction sync is enabled, it makes more sense to me if this sync-date partial is only included if transaction sync is enabled.

Take the case where transaction sync feature was enabled, but user no longer wishes to use it and so disables it. Regardless of whether transaction sync is enabled, the TJ sync date will still be displayed (i.e. the .phtml template/partial will be applied to the page) if the order has a tj_salestax_sync_date value stored on it.

The argument could be made that because the data exists, it should be visible to admin, but I'm not sure I agree. If after installing the TJ SalesTax module, you decide to uninstall it, clear static content, recompile, and run upgrade, the tj_salestax_sync_date column will still exist on the sales_order table, and any existing values will remain. That data is still there on the order, just hidden because TJ is "turned off" in M2. While that is partly due to the fact that our module doesn't include extension attributes properly for legacy reasons, I think it is also indicative of the fact that whether or not we include this UI template should first depend on if our module is enabled, then if a feature is configured, and finally the state of an order. The state of the order can then affect what we display within that UI template.

If we used extension attributes as recommended instead of storing our values directly on the Magento_Sales tables/models, would we still load an order's tj_salestax_sync_date value in our extension data when the TaxJar transaction sync feature flag is disabled?

What are your thoughts?

dallendalton: I'll think about this more but for now I think its probably fine as is. If we do decide to make this change later we can open up a new card for it.

dallendalton (approved):

jordan-k-johnson (approved):
LGTM

PR #236, opened 2022-01-19
XC-7: Sales order calculation status

sethobey:
### Context

Refactors TaxJar Sales Order view additional info.

### Description

- Sales Order TaxJar Info's section "Sync Status" displays one of three potential messages depending on order status and sync status
- Sales Order TaxJar Info's section "Calculation Status" displays one of three potential messages depending on order extension metadata's `tax_result` value.
  - Order and Creditmemo calculation status will continue to be displayed in sales_order_view and sales_order_creditimemo_view as it has been historically.
- To align with recent changes to similar feature implementation in WooCommerce plugin, PR does not include "Advanced" section. Order Metadata does not store JSON request/response body.

### Performance

Performance impact should be negligible, but feature adds an additional DB write operation for the order extension data.

#### Tax not calculated
1. Checkout as Customer with TJ extension disabled
2. Enable TaxJar extension
3. Navigate to sales order view as Admin
4. Observe calculation status

image

#### Tax calculate success
1. Checkout as Customer with TJ extension enabled for checkout
2. Navigate to sales order view as Admin
3. Click "TaxJar Information" tab
4. Observe calculation status

image

#### Tax calculate error
1. Simulate API failure. e.g. modify code in `SmartCalcs.php` to throw exception on API call
2. Checkout as Customer with TJ extension enabled for checkout
3. Navigate to sales order view as Admin
4. Click "TaxJar Information" tab
5. Observe calculation status

image

#### Sync status feature disabled
1. Navigate to sales order view as Admin with TJ extension transaction sync disabled
2. Click "TaxJar Information" tab
3. Observe sync status

image

#### Sync status awaiting
1. Checkout as Customer with TJ extension transaction sync enabled
2. Navigate to sales order view as Admin
3. Click "TaxJar Information" tab
4. Observe sync status image
5. Invoice order image

#### Sync status success
1. Using previous order, ship order to move status to complete
2. Click "TaxJar Information" tab
3. Observe sync status

image

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x


dallendalton (changes requested):
Everything I tested looked good. However, I think we might want to make some improvements. Specifically in Model/Smartcalcs.php in the getTaxForOrder method. Everywhere where this method short circuits and returns (due to no nexus, etc.) we aren't actually performing a tax calculation request and are just falling back to Magento rates. When an order is created from one of these quotes, the calculation status message is "Tax was calculated in realtime through the TaxJar API." which is not really true for the order. 

Most often when customers think an order should have had tax calculated and it didn't it wasn't due to an unknown exception (which this PR currently handles) but rather due to our validation that short circuits the method. 

I think we need to surface the reason why we didn't calculate tax and that we reverted to Magento default tax calculations when this occurs in the metadata.

sethobey:
> Everything I tested looked good. However, I think we might want to make some improvements. Specifically in Model/Smartcalcs.php in the getTaxForOrder method. Everywhere where this method short circuits and returns (due to no nexus, etc.) we aren't actually performing a tax calculation request and are just falling back to Magento rates. When an order is created from one of these quotes, the calculation status message is "Tax was calculated in realtime through the TaxJar API." which is not really true for the order.
> 
> Most often when customers think an order should have had tax calculated and it didn't it wasn't due to an unknown exception (which this PR currently handles) but rather due to our validation that short circuits the method.
> 
> I think we need to surface the reason why we didn't calculate tax and that we reverted to Magento default tax calculations when this occurs in the metadata.

Hey @dallendalton,
Good call out there. I've updated Smartcalcs request validation and now we should show a message like so:

image

jordan-k-johnson (approved):
LGTM

Model/Smartcalcs.php
```diff
@@ -543,21 +576,24 @@ private function _getLineItems(
...
-                'id' => $nexusAddress->getId(),
-                'country' => $nexusAddress->getCountryId(),
-                'zip' => $nexusAddress->getPostcode(),
-                'state' => $nexusAddress->getRegionCode(),
-                'city' => $nexusAddress->getCity(),
-                'street' => $nexusAddress->getStreet()
-            ];
-        }
+        $nexusCollection = $this->nexusFactory->create()->getCollection()->addStoreFilter($storeId);
+        return array_map([$this, '_getNexusAddress'], $nexusCollection);
```
dallendalton: I'm getting an error here:
[2022-01-31 17:31:28] main.CRITICAL: Report ID: webapi-61f81cf011158; Message: Warning: array_map(): Expected parameter 2 to be an array, object given in /var/www/html/app/code/Taxjar/SalesTax/Model/Smartcalcs.php on line 580 {"exception":"[object] (Exception(code: 0): Report ID: webapi-61f81cf011158; Message: Warning: array_map(): Expected parameter 2 to be an array, object given in /var/www/html/app/code/Taxjar/SalesTax/Model/Smartcalcs.php on line 580 at /var/www/html/vendor/magento/framework/Webapi/ErrorProcessor.php:208, Exception(code: 0): Warning: array_map(): Expected parameter 2 to be an array, object given in /var/www/html/app/code/Taxjar/SalesTax/Model/Smartcalcs.php on line 580 at /var/www/html/vendor/magento/framework/App/ErrorHandler.php:61)"} []

dallendalton (changes requested):

Model/Smartcalcs.php
sethobey: 🤦 I believe this was due to a bad "soft" reset.. I've abstracted these nexus-related operations into a separate Helper class and added test coverage in `\Taxjar\SalesTax\Test\Unit\Helper\NexusTest.php` to prevent regression.

sethobey:
Also worth noting, since opening this PR, I've discovered some differences in how `Magento_Sales` implements extension attributes vs. other core-modules' extensible interfaces.

The addition of a `created_at` column to our metadata table can lead to an ambiguous `where`-clause in some situations. While this could be mitigated in other ways, such as `di.xml` configuration or use of a plugin, it seems more practical to remove the offending column since it is not necessary rather than implement additional code.

Helper/Nexus.php
```diff
@@ -0,0 +1,109 @@
...
+     * Get nexus collection formatted as an array of address data arrays
+     *
+     * @param string|int|null $storeId optionally filter result by store_id
+     * @return array
+     */
+    public function getNexusAddresses($storeId): array
+    {
+        /** @var array|\Taxjar\SalesTax\Api\Data\Tax\NexusInterface[] $nexusArray */
+        $nexusArray = $this->getNexusCollection($storeId)->getItems();
+        return array_map(fn ($nexus) => $this->getNexusData($nexus), $nexusArray);
```
dallendalton, line 46: When making a request to v2/taxes I am getting a 406 response with this body:

{"status":"406","error":"Not Acceptable","detail":"nexus_addresses is invalid"}

The nexus_addresses parameter for that request contains this data:

{"42":{"id":"42","country":"US","zip":"84651","state":"UT","city":"Payson","street":"123 Main St"}}

I believe that having it start at index 42 may be the problem (haven't tested that though) and that it comes from the output of this method.

dallendalton (changes requested):

Helper/Nexus.php
sethobey: Fixed and updated unit tests to account for this

dallendalton (approved):

PR #312, opened 2022-04-18
fix: Config scoped address validation

sethobey:
### Context

EU-based merchant with US storeview observed an issue with address validation. Address validation feature was disabled globally, but should've been enabled for the US storeview's scope.

### Description

Adds request parameter value for address validation scope configuration. Address validation requests in checkout will pass the current store code to the backend which is used to find the current store scope ID in a WebAPI request.

Customer address validation in Adminhtml will still defer to the global scope's configuration value.

### Performance

Bug fix

### Testing

1. Observed successful checkout address validation for store with address validation enabled while global address validation disabled.
2. Observed bypassed checkout address validation for store with disabled global address validation config
3. Observed bypassed customer address validation in adminhtml due to feature being disabled in global config 

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3

- [ ] Magento Commerce (EE)
- [ ] Magento B2B
- [ ] Magento Cloud

- [X] PHP 7.x

### DIFF
--- a/Api/AddressValidationInterface.php
+++ b/Api/AddressValidationInterface.php
@@ -28,6 +28,7 @@ interface AddressValidationInterface
      * @param string $region
      * @param string $country
      * @param string $postcode
+     * @param string $storeCode
      * @return mixed
      */
     public function validateAddress(
@@ -36,6 +37,7 @@ interface AddressValidationInterface
         $city = null,
         $region = null,
         $country = null,
-        $postcode = null
+        $postcode = null,
+        $storeCode = null
     );
 }
--- a/Model/AddressValidation.php
+++ b/Model/AddressValidation.php
@@ -26,6 +26,8 @@ use Magento\Framework\App\CacheInterface;
 use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\Exception\LocalizedException;
 use Magento\Store\Model\ScopeInterface;
+use Magento\Store\Model\Store;
+use Magento\Store\Model\StoreManagerInterface;
 use Taxjar\SalesTax\Api\AddressValidationInterface;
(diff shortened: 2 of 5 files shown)


Citizen4our:
Hello
also [here](https://github.com/taxjar/taxjar-magento2-extension/blob/c8fbb14de720128ba15c0f4c7acdc6b28dc90dce/view/base/web/js/model/address_validation_core.js#L30) needed to use 'mage/url' for build url, because some magento instances might be using 'example.com/us/' in US store for example

sethobey:
> Hello also [here](https://github.com/taxjar/taxjar-magento2-extension/blob/c8fbb14de720128ba15c0f4c7acdc6b28dc90dce/view/base/web/js/model/address_validation_core.js#L30) needed to use 'mage/url' for build url, because some magento instances might be using 'example.com/us/' in US store for example

Hi @Citizen4our, this pull request is for addressing a specific internal issue which as far as I know, did not involve any issues with base url.

For what it's worth, I think I understand what you are suggesting, but it isn't clear to me how that would be implemented here.

Our address validation core is already in use in the frontend and backend admin areas with the same WebAPI URI. I've only added an optional `store_name` parameter here for frontend address validation to allow scoping frontend requests and enabling/disabling address validation for individual scopes which should work regardless of store URL structure.

I considered passing the URI to the address validation core component, but that would've resulted in greater refactor which is undesirable since we've just recently squashed numerous outstanding JS-related address validation bugs.

If this hasn't _addressed_ (😄) your concern, please feel free to [open a new issue here](https://github.com/taxjar/taxjar-magento2-extension/issues/new) to describe the request or change you're suggesting in greater detail.

Thanks!

dallendalton (approved):

jordan-k-johnson (approved):

Issue #375, opened 2024-06-11
Unable to save customers with certain TJ fields null


snoop0x7b:
### Prerequisites

- [X] Are you running the latest version of our extension?
- [X] Have you reviewed our [integration guide](https://developers.taxjar.com/integrations/guides/magento2/)?
- [X] Did you search through [existing issues](https://github.com/taxjar/taxjar-magento2-extension/issues)?

### Description

When saving customers if the tj_exemption_regions or tj_last_sync are null, we get the following PHP errors. This occcurs on Magento 2.4.6-P5 and PHP 8.2. This did not occur on PHP7.4 for us. 

 Call to a member function getValue() on null at vendor/taxjar/module-taxjar/Observer/Customer/Save.php:57

Call to a member function getValue() on null in vendor/taxjar/module-taxjar/Observer/Customer/Save.php:80

### Steps to Reproduce
1. Go edit a customer
2. Make sure the last sync date and exempt regions are null.
3. Save the customer

### Expected Result
Editing the customer should be successful

### Actual Result
PHP error - customer does not save.

### Versions
Magento 2.4.6-p5
- [X] Magento 2.4
- [ ] Magento 2.3

- [x] Magento Open Source (fka Magento 2 Community Edition)
- [X] Adobe Commerce (fka Magento 2 Enterprise Edition)

- [X] PHP 8.2
- [x] PHP 8.1
- [ ] PHP 7.4
- [ ] PHP 7.3

snoop0x7b:
Looks like @smolentzov-stripe fixed this issue.

PR #376, opened 2024-06-11
Fix issue related to saving customers

snoop0x7b:
### Context
This pull request fixes issue #375 

### Description
This fixes the bug in issue #375 

### Performance
No impact on performance - It adds null checks. 

### Testing

You can test this by replicating the issue in #375 and then applying this patch and seeing if it still causes the issue. After doing these changes, we stopped having that particular issue and were again able to save customer accounts as expected. 

#### Versions

- [X] Magento 2.4
- [ ] Magento 2.3

- [ ] Magento Open Source (formerly Magento 2 Community Edition)
- [X] Adobe Commerce (formerly Magento 2 Enterprise Edition)

- [X] PHP 8.2
- [ ] PHP 8.1
- [ ] PHP 7.4
- [ ] PHP 7.3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-## [4.0.0] - 2025-08-25
+## [4.0.1] - 2025-08-25
+### Fixed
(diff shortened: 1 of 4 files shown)


snoop0x7b:
Hey Taxjar team

Any plans to merge this PR? Your extension may not work correctly with supported versions of Magento without this update. We're currently relying on a patch I created in our project.

PR #383, opened 2025-04-24
Version 3.0.0

smolentzov-stripe:
Version bump. The prior release was backwards incompatible past 2.4.6 and should have been a major version bump, this corrects that.


culley-stripe (approved):
🎉

